PR #18, opened 2017-12-04
Fix exclude vendor folder

bestplay:
Think about :
vendor/xxx.go
xxx/vendor/xxx.go


easonlin404:
LGTM. Thanks all for your contributions. 😄

PR #27, opened 2018-01-10
add more golang build-in Numeric types support by swagger.

xykong:
add type support by golang and swagger

+// reference: https://swagger.io/docs/specification/data-models/data-types/
+// reference: https://golang.org/ref/spec#Numeric_types


xykong:
The code is mismatch with go import path, I'll create new one.

PR #37, opened 2018-01-24
Transfer float32 and float64 to number type

codebear4:


easonlin404:
@codebear4 why do you want to transfer float32 and float64 to number type?

codebear4:
Now the `swag init` command generate `float64` field as same as it is in the `docs.go`, then the Swagger UI parse the `float64` type as "Unknown Type: float64"。So I think the `number` type should be fit for the `float64` and `float32` types in Golang.
snipaste_2018-01-24_20-01-05

PR #49, opened 2018-02-22
add array of standard types

nadilas:
cleanup array of structs method and add array of standard types (e.g. []string)

add fix TransToValidSchemeType(typeName string) to support user defined types in ast.Ident fields

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -299,7 +299,11 @@ func (parser *Parser) parseAnonymousField(pkgName string, field *ast.Field, prop
 
 func (parser *Parser) parseField(field *ast.Field) (propName, schemaType string, arrayType string) {
 	schType, arrType := getPropertyName(field)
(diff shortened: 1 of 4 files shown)


nadilas:
Found one additional line, which needs to be swapped:

`parser.go:294`:
```go
CheckSchemaType(schType)
```

should be:
```go
if len(arrType) == 0 {
	CheckSchemaType(schType)
} else {
       CheckSchemaType("array")
}
```

easonlin404:
@nadilas can you help to fix testing?

```
--- FAIL: TestParseSimpleApi (0.02s)
```

nadilas:
Yes, sorry I missed that.

easonlin404:
@nadilas  LGTM. Thanks for all your contributions. 💯

PR #50, opened 2018-02-24
return json from build to enable usage elsewhere

nadilas:
possible usage e.g.: cli param -printJson to print doc.json to stdout instead of serving via http


easonlin404:
LGTM

Issue #17, opened 2017-11-30
Support parsing interface type in swagger schema object


easonlin404:

easonlin404:
ref:
https://github.com/swaggo/swag/issues/16

pei0804:
memo: property.go: 44
I should add interface pattern.

PR #74, opened 2018-03-24
feat: add support parsing interface type in swagger schema object (#17)

pei0804:
close https://github.com/swaggo/swag/issues/17


easonlin404:
LGTM

PR #91, opened 2018-04-08
Fix ignore json tag option

pei0804:
Please first merge  https://github.com/swaggo/swag/pull/89

ignore json tag option

```go
type Model struct {
    Field int `json:"field,omitempty"`  -> field
}
```


pei0804:
ok

PR #100, opened 2018-04-19
update: pull out common splitting comment

cia-rana:
I pulled out the common statement `strings.Split(comment.Text(), "\n")` in parser.go.


easonlin404:
@cia-rana thanks for your contribution. 💯

Issue #109, opened 2018-05-04
Special type and cross package support


tx991020:
http://oq1xuulg6.bkt.clouddn.com/97c3b42fbc5bebf327293a31f98a0204.jpg

pei0804:
How do you have struct field?

tx991020:
dao.Product  like this   
![](http://oq1xuulg6.bkt.clouddn.com/dd149276c1af8e95093b0d7291cd0d28.jpg)

tx991020:
http://oq1xuulg6.bkt.clouddn.com/dd149276c1af8e95093b0d7291cd0d28.jpg

pei0804:
What is JSON TIME？
If it is a library, please give me a URL.

tx991020:
@pei0804                    type JSONTime struct {T time.Time}

pei0804:
I see.

easonlin404:
@tx991020 please reopen it if you have any problems.

tx991020:
@easonlin404  @pei0804   the bug already exist  type JSONTime struct {T time.Time}
http://oq1xuulg6.bkt.clouddn.com/b0f6c8416cd2a208c58b0aa889120696.jpg
http://oq1xuulg6.bkt.clouddn.com/19aad5d4ac395939503e0978509a2956.jpg

pei0804:
Did you finished update?

tx991020:
@pei0804 @easonlin404  I have finished update  . the bug already exist

pei0804:
I'll Check.

edmundluk3:
`a.go`
```
package a
struct user {
  name string
}
```

`b.go`
```
package b
struct response {
admin user // contains another here
}
```

```
package b
// @Tags post
// @Accept json
// @Produce json
// @Success 200 {object} response
// @Router /post/ [get]
```

will above code be the case supported???

pei0804:
sorry.
cross package import not support.

pei0804:
@tx991020
You use the type prepared by the standard. So, I'm creating plugin for same like issue.

tx991020:
panic: not supported 'astSelectorExpr' yet.

zhongkeqin:
1. I see that `ParseAPI` method have parsed all user define struct type in project, so I change the `parseStruct` method to support cross package imports: If cannot find pkgName.typeName in `parser.TypeDefinitions` then ignore pkgName and use `typeName` to search
```
[...]

PR #142, opened 2018-06-04
Added Support for Cross Package Imports

paragtokopedia:
**Describe the PR**
Added Cross Package Support for Struct

**Relation issue**
https://github.com/swaggo/swag/issues/109

**Additional context**
if a struct is depending on other struct from different package, it is not shown in reference.


paragtokopedia:
if a struct depends on cross package struct then arrayType returned looks like 
&{packageName structField}, i removed &{ and } and split it on space to get packageName and structField, before this parser.TypeDefinitions[pkgName][structField.arrayType] was not able to find definition because both pkgName and structField.arrayType were wrong (pkgName was current package and not cross package name). Now it is working fine 

i tested for following use case of mine 

type VCCAddUserRes struct {
	AlreadyExists    []db_model.VCCUser `json:"already_exists"`
	Invalid          []db_model.VCCUser `json:"invalid"`
	TotalSuccessFull int                `json:"total_success_full"`
}

here db_model.VCCUser is cross package and structField.arrayType value for this is &{db_model VCCUser}

pei0804:
Can you add test?

PR #152, opened 2018-06-14
feat: support description with newline

hykuan:
swag will treat /n as /(92) n(110) not LF(10), so i replace it.

and i use gopkg.in/yaml.v2 which can generated correct yaml format with newline


hykuan:
what should i do to make CI pass ?

example/basic/docs/docs.go
```diff
@@ -188,13 +188,13 @@ var doc = `{
         "web.APIError": {
             "type": "object",
             "properties": {
-                "CreatedAt": {
+                "createdAt": {
```
easonlin404: Cannot change to Lower case begin.

example/basic/docs/docs.go
```diff
@@ -188,13 +188,13 @@ var doc = `{
...
                     "type": "string"
                 },
-                "ErrorCode": {
+                "errorCode": {
```
easonlin404: Same as above

example/basic/docs/docs.go
```diff
@@ -188,13 +188,13 @@ var doc = `{
...
                     "type": "integer"
                 },
-                "ErrorMessage": {
+                "errorMessage": {
```
easonlin404: Same as above

hykuan:
OK, you can close the PR, and i will wait someone implements the feature.

easonlin404:
@hykuan PR is welcome.

PR #163, opened 2018-07-04
feat: support empty response body when only have status code

### LINKED ISSUE #161: A way to specify an empty/nil response body
My api for 404 responses just returns a 404 status code, and for deletes just returns a 204 status code. These responses have no body, and I would like to keep that in the documentation.
A way to indicate the status code with no additional response types. 
Something like `// @Failure 404 nil` ? 

not document the failure type 

Thanks for the library! it's pretty dope.

easonlin404:
**Describe the PR**
fix #161


pei0804:
LGTM

PR #167, opened 2018-07-12
after modify docs no need build project always

wksw:
**Describe the PR**
e.g. after modify docs no need build project always.

**Relation issue**
e.g. 

**Additional context**
when modify api comments need build project

### DIFF
--- a/gen/gen.go
+++ b/gen/gen.go
@@ -50,11 +50,13 @@ func (g *Gen) Build(searchDir, mainAPIFile, swaggerConfDir, propNamingStrategy s
 	swaggerYAML.Write(y)
 
 	packageTemplate.Execute(docs, struct {
-		Timestamp time.Time
-		Doc       string
+		Timestamp   time.Time
+		Doc         string
+		SwaggerJSON string
 	}{
-		Timestamp: time.Now(),
-		Doc:       "`" + string(b) + "`",
+		Timestamp:   time.Now(),
+		Doc:         "`" + string(b) + "`",
+		SwaggerJSON: path.Join(swaggerConfDir, "swagger.json"),
 	})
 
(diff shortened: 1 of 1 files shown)


easonlin404:
@wksw lint check failed, please fix by executed [gofmt](https://golang.org/cmd/gofmt/)  cmd.

```
/home/travis/gopath/src/github.com/swaggo/swag/gen/gen.go
^ - Repo contains improperly formatted go files
```

gen/gen.go
```diff
@@ -69,14 +71,26 @@ package docs
...
+	"io/ioutil"
 )
 
 var doc = {{.Doc}}
 
 type s struct{}
 
 func (s *s) ReadDoc() string {
-	return doc
+	file, err := os.Open("{{.SwaggerJSON}}")
```
easonlin404: @wksw would you tell us why there is using `{{.SwaggerJSON}}`? It seems like will go wrong .

wksw: use `{{.SwaggerJSON}}` it  means read from file not from variable `doc`(line 78)
the variable doc's content is same with the file `{{.SwaggerJSON}}`

easonlin404: yep, I known `{{.SwaggerJSON}}` but that cause `variable doc(line 78)` useless.

wksw: got it, you means it will cause `doc defined but not used` error,  right???
when i build my project this error not occurred, i will check it carefully

easonlin404: no, I mean  `variable doc(line 78)` now don't be used. Because have using `{{.SwaggerJSON}}`instead of `variable doc(line 78)`.

easonlin404:
please create the new PR for it because toward master branch so far away.

Issue #159, opened 2018-06-29
Support nested imports for Struct definitions


inoda:
**Is your feature request related to a problem? Please describe.**
I have a Struct:
```
type RevenueMetricsPayload struct {
	RevenueMetrics procedures.RevenueMetricsTimeline `json:"revenueMetrics"`
	Counts         []count                           `json:"counts"`
	ProcessedAt    string                            `json:"processedAt"`
}
```

When generating docs, I get this message 

> RevenueMetricsTimeline is not supported. but it will be set with string temporary

**Describe the solution you'd like**
I would like the nested import to be followed, and have `procedures.RevenueMetricsTimeline` in my example built out based on the defined Struct.

**Describe alternatives you've considered**
I could manually override the generated documentation, or write a script to do it, or create another struct just for documentation purposes.

easonlin404:
@inoda PR is welcome 😄

pei0804:
I support this.

PR #171, opened 2018-07-22
add support cross package

pei0804:
**Describe the PR**
I also made it possible to create a model in Cross package

**Relation issue**
close https://github.com/swaggo/swag/issues/159

**Additional context**
Very bad code, but it works.

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -441,12 +441,12 @@ type structField struct {
 	arrayType    string
 	formatType   string
 	isRequired   bool
+	crossPkg     string
 	exampleValue interface{}
 }
 
 func (parser *Parser) parseStruct(pkgName string, field *ast.Field) (properties map[string]spec.Schema) {
 	properties = map[string]spec.Schema{}
-	// name, schemaType, arrayType, formatType, exampleValue :=
 	structField := parser.parseField(field)
 	if structField.name == "" {
 		return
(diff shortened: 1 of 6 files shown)


easonlin404:
@pei0804 can you provide an example about how to enable cross pkg?

pei0804:
It's here.
https://github.com/swaggo/swag/blob/0a99c8c47b9d78dccaec2d4031d8978f08e5cff6/testdata/simple/web/handler.go#L64

easonlin404:
IMO seems like can make code more concise, rest LGTM.

pei0804:
I think so too.

t-evers:
I can confirm cross-package works now. But I continue to run into the "not supported, but it will be set with string temporary when it comes to using models from another project.
Example layout as it resides in go/src:
our.domain/project1/models/somedata.go
our.domain/project2/models/otherdata.go

Now, if project2 has the REST API to be documented and the models in "otherdata.go" contain types declared in project1/somedata.go, I ran into that problem.

Can you confirm that and will that be fixed too?

pei0804:
I want a fail code. Can you give?

t-evers:
Of course. Here you go. 
Output of "swag init":
[example.zip](https://github.com/swaggo/swag/files/2235530/example.zip)
2018/07/27 13:02:55 Generate swagger docs....
2018/07/27 13:02:55 Generate general API Info
SomeData is not supported. but it will be set with string temporary. Please report any problems.2018/07/27 13:02:55 Generating api.ResponseObject
2018/07/27 13:02:55 create docs.go at  docs/docs.go

pei0804:
thx. I'll check tommrow.

pei0804:
I understood the problem.
swag is parsing directories under the location where init was executed.

```
our.domain/project1/models/somedata.go
our.domain/project2/models/otherdata.go
```

Project2 can't be analyzed because project 1 is'nt visible.
It seems that it is necessary to put project1 on project 2 or less now.

pei0804:
@t-evers 
for you
https://github.com/swaggo/swag/pull/179#pullrequestreview-142388523

Issue #186, opened 2018-08-16
generated model field descriptions always ends with "\n"


abdulito:
For example
```
type Service struct{      
        // Service ID
	ID string `json:"id"`
}

```

will generate

```
 "v1.Service": {
            "type": "object",
            "properties": {
                "id": {
                    "description": "Service ID\n",
                    "type": "string"
                },
```

Thanks!

pei0804:
https://github.com/swaggo/swag/pull/187

PR #187, opened 2018-08-16
#186: remove new line from struct field description

abdulito:
This is for trimming new lines from struct field descriptions

**Relation issue**
https://github.com/swaggo/swag/issues/186


pei0804:
thx

Issue #188, opened 2018-08-19
Responses cannot reference types in vendored external packages


dselans:
Thank you for starting this project. This is the *exact* same thing I wanted to do w/ `yvasiyarov/swagger`, but unfortunately did not have enough time. It seems that you and a number of contributors got the ball rolling in the right direction with this project. Awesome and well done!

**Describe the bug**
When attempting to generate docs that contain a response that references a type in a vendored package, `swag` will run into a `can not parse response comment` (which actually is `package was nil`, that comes from `operation.findTypeDef(...)`).

Example:

We have spec comments that look like this:

```
// @Summary Fetch a request object by ID
// @Description Use this endpoint to determine the state of the request
// @Tags request
// @Accept json
// @Produce json
// @Param request_id path string true "Request ID UUID"
// @Success 200 {object} types.Request "Request has completed successfully"
// @Success 201 {object} types.Request "Request is in-progress"
// @Success 202 {object} types.Request "Request has been queued"
// @Failure 412 {object} rye.JSONStatus "Request ran into an error"
// @Failure 500 {object} rye.JSONStatus "Unexpected server error"
// @Router /v1/request/{request_id} [get]
```

The important bit is the `// @Failure 412 {object} rye.JSONStatus "Request ran into an error"` which is referencing the `rye` package, which is external _and vendored_.

This is what you'll see when you attempt to generate docs:

```
$ go run /Users/dselans/Code/go/src/github.com/swaggo/swag/cmd/swag/main.go init
2018/08/18 21:52:16 Generate swagger docs....
2018/08/18 21:52:16 Generate general API Info
2018/08/18 21:52:17 ParseComment panic:can not parse response comment "412 {object} rye.JSONStatus "Request ran into an error""
panic: ParseComment panic:can not parse response comment "412 {object} rye.JSONStatus "Request ran into an error""

goroutine 1 [running]:
log.Panicf(0x14b76b3, 0x16, 0xc42018d500, 0x1, 0x1)
	/usr/local/go/src/log/log.go:333 +0xda
github.com/swaggo/swag.(*Parser).ParseRouterAPIInfo(0xc4203da480, 0xc4204b6f80)
	/Users/dselans/Code/go/src/github.com/swaggo/swag/parser.go:294 +0x1da
github.com/swaggo/swag.(*Parser).ParseAPI(0xc4203da480, 0x14b0909, 0x2, 0x14b1ae9, 0x7)
[...]

PR #189, opened 2018-08-19
minor fix to address referencing types in external, vendored pkgs #188

dselans:
**Describe the PR**
If the pkg referenced in `importPath` is actually a package that is vendored, use the "full path" for the `program.Package(..)` call. Calling on `program.Package(originalmportPath)` will cause the return pkg to be nil, even though the `.Load(..)` succeeded.

**Relation issue**
#188 

**Additional context**
I added some extra new lines during debugging - sorry! Looking at it now, I suppose the map iteration can happen after initial `program.Package(..)` call fails.. but that may be a overkill.

### DIFF
--- a/operation.go
+++ b/operation.go
@@ -434,22 +434,36 @@ func findTypeDef(importPath, typeName string) (*ast.TypeSpec, error) {
 	if err != nil {
 		return nil, err
 	}
+
 	conf := loader.Config{
 		ParserMode: goparser.SpuriousErrors,
 		Cwd:        cwd,
 	}
+
 	conf.Import(importPath)
+
 	lprog, err := conf.Load()
 	if err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


dselans:
Sidenote: after codefix is applied, you'll see this output instead:

```
$ go run /Users/dselans/Code/go/src/github.com/swaggo/swag/cmd/swag/main.go init
2018/08/18 22:58:53 Generate swagger docs....
2018/08/18 22:58:53 Generate general API Info
2018/08/18 22:58:55 Generating types.Action
2018/08/18 22:58:55 Generating types.RequestPayload
2018/08/18 22:58:55 Generating types.Request
2018/08/18 22:58:55 Generating rye.JSONStatus
2018/08/18 22:58:55 create docs.go at  docs/docs.go
```

pei0804:
Oh Thx!!!!!!!!!

PR #192, opened 2018-08-22
Fixbug: Panic when parse param with object type

collinmsn:
swag init panic with message:

panic: assignment to entry in nil map


pei0804:
Can you add a test?

Around here add nil map.
https://github.com/swaggo/swag/blob/542fa3699f6e267a876a42c356172c73fc04b062/testdata/simple/web/handler.go#L34

bendoerr:
fwiw, tested rebased onto master, does fix the problem

Issue #164, opened 2018-07-07
Support custome type field


pei0804:
for https://github.com/swaggo/swag/issues/158

https://github.com/swaggo/swag/issues/158

pei0804:
## 1. Add test for `getPropertyName`
https://github.com/swaggo/swag/blob/015cc4041d62df0b61cb051d96a3820dd9fb0d65/property.go#L42

## 2. Add cmd option for `customeType`

```console
$ swag init --customeType userDifine.so
```

`.so`  is  plugin prepared by the standard package of go.  
 [example](https://github.com/pei0804/go-plugin-sample/tree/master/basic)  

### How to customeType.

1. Define type map.

```go
var customeTypeMap = map[string]string{
	"customeField": "string",
	"customeField2": "number",
}

var customePkgAndTypeMap = map[string]string{
	"pkg.customeField": "string",
	"pkg2.customeField": "number",
}
```

2. Generate `.so`  and do swag init

```console
$ go build -buildmode=plugin -o typeDefine.so typeDefine.go
$ swag init --customeType userDifine.so
```

pei0804:
This approach was useless. I added only the test for property.

pei0804:
I want to resolve the cross package before doing this. It is bottleneck.

superyfwy:
-buildmode=plugin not supported on darwin/amd64

pei0804:
> -buildmode=plugin not supported on darwin/amd64

Oh

pei0804:
i'm planning for custom type solution with the following code.

```go
package main

import (
	"fmt"
	"reflect"
	"time"
)

type Investment struct {
	Price  float64
	Str    *A
	Map    B
	Time   time.Time
	Rating int64
}

type A string
type B map[string]string

func main() {
	a := A("A")
	inv := Investment{Price: 534.432, Str: &a, Rating: 4, Map: map[string]string{}, Time: time.Now()}
	s := reflect.ValueOf(&inv).Elem()
	variableParms := make([]interface{}, s.NumField())
	for i := 0; i < s.NumField(); i++ {
		variableParms[i] = s.Field(i).Addr().Interface()
	}

	mockFunction(inv.Price, inv.Str, inv.Rating, inv.Map, inv.Time)
}

func mockFunction(values ...interface{}) {
	for i, value := range values {
		rv := reflect.ValueOf(value)
		fmt.Printf("value %d has Type: %s and Kind %s\n", i, rv.Type(), rv.Kind())
[...]

PR #198, opened 2018-08-31
Custom primitive types

abdulito:
Handle custom primitive types

https://github.com/swaggo/swag/issues/164


duguying:
how to use

duguying:
it seems not support `type Time time.Time`

PR #206, opened 2018-09-12
refactor: enhance error handle

easonlin404:
Enhance error handle, there are a lot of  snippets should  returns error instead of panic.


pei0804:
Good!

PR #212, opened 2018-09-20
fix: missing schema convert

elvizlai:
**Describe the PR**
fix missing schema convert

ex:

```
type Gender int32

type User struct {
	Name string
	Sex  Gender
}
```

the generated doc should be `sex integer` but not `sex int32`


pei0804:
Thx.

PR #228, opened 2018-10-24
stop supporting go 1.8.x

rubensayshi:
**Describe the PR**
remove travis testing go 1.8.x support.

go 1.8.x is no longer considered "actively maintaned" (see https://github.com/golang/go/wiki/Go-Release-Cycle) and as a result golint has started using some go 1.9 features and can no longer be installed in a go 1.8.x env.

this is currently resulting in breaking all other PR builds.

**Relation issue**
https://github.com/golang/go/issues/28291


easonlin404:
LGTM

Issue #191, opened 2018-08-22
Multiline description


bpsnifer:
Can you add feature: when repeating the comment tag @description, add it with a new line. E.g:
```go
    case "@description":
    if parser.swagger.Info.Description != "" {
        parser.swagger.Info.Description += "\n"
    }
    parser.swagger.Info.Description += strings.TrimSpace(commentLine[len(attribute):])
```
```go
    case "@description":
        if operation.Description != "" {
            operation.Description += "\n"
        }
        operation.Description += lineRemainder
```

pei0804:
What @easonlin404  you think?

easonlin404:
Looks like a good catch😀

pei0804:
Welcome PR.

lfaoro:
awesome, with this we could have multiline descriptions! although is there a way to achieve multiline descriptions without repeating the key?

e.g. this is what one would do in yaml

anarcher:
How about to add `@import`  annotation for long description  ?  :-)

PR #229, opened 2018-10-25
Add multiline description for main parser

kaythxbye:
**Describe the PR**
Like already discussed (and implemented) in #191 for single operations, the main description does not support multiline descriptions yet. I added the code from the PR that was created from that issue and implemented that functionality in the parser.go

**Relation issue**
#191 

**Additional context**
Markdown Headings do not work as far as I tested.


easonlin404:
fixed via https://github.com/swaggo/swag/pull/278

PR #232, opened 2018-11-01
Translate array type into a valid swagger type

nkly:
First of all, thank you very much for the wonderful project :+1: 

Now to the issue. Previously, the following structure:

    type Foo struct {
        Bar []int
    }

would generate this swagger spec:

    ...
    Bar:
      items:
        type: int
    ...

`int`, however, is not a valid swagger type. It is inserted because array types are not processed by the translation function. This commit applies the type translation function to all the places where array types are used.

### DIFF
--- a/parser_test.go
+++ b/parser_test.go
@@ -551,6 +551,12 @@ func TestParseSimpleApi(t *testing.T) {
                     "format": "int64",
                     "example": 1
                 },
+                "int_array": {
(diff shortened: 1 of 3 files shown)


easonlin404:
@nkly Thanks for your contributions. 💯

nkly:
@easonlin404 Thank you for the fast review :+1:

PR #234, opened 2018-11-02
Allow overriding swagger type in struct fields

nkly:
This PR adds support for a new struct field tag, `swaggertype`. It allows user to redefine the resulting swagger type of a struct field. It comes in 2 formats:

- `swaggertype:<prim_type>` sets the swagger type of the field to the given `<prim_type>`
- `swaggertype:array,<prim_type>` sets the swagger type of the field to "array" and the swagger type of its items to `<prim_type>`

This feature is useful when you have custom types that are not supported by `swag` tool yet, but nevertheless have a corresponding swagger type that fits them well, i.e. `sql.NullInt64` can be simply `integer`, `big.Float` can be `number`, etc. While it would be nice to support these types out of the box, it is not possible to do so for every custom wrapper; that's why I think this feature could be valuable, but please feel free to disagree :)


pei0804:
Oh nice idea.
I thought it was good.

PR #239, opened 2018-11-05
Add support for the attributes in struct fields

nkly:
This commit adds support for the following attributes in the struct fields:

- `minimum`, for numeric fields, via `minimum:"value"` struct tag
- `maximum`, for numeric fields, via `maximum:"value"` struct tag
- `minLength`, for string fields, via `minLength:"value"` struct tag
- `maxLength`, for string fields, via `maxLength:"value"` struct tag
- `enums`, via `enums:"value1,value2,value3"` struct tag
- `default`, via `default:"value"` struct tag

All these tags except `default` are also supported in array fields.

### DIFF
--- a/README.md
+++ b/README.md
@@ -432,6 +432,16 @@ Make it AND condition
 // @Param default query string false "string default" default(A)
 ```
 
+It also works for the struct fields:
+
+```go
+type Foo struct {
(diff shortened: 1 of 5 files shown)


easonlin404:
@nkly could you help resolve conflicting files, rest LGTM?

Issue #221, opened 2018-09-29
Nested struct field from different package not documented properly


bojand:
**Describe the bug**

If we have code like this:

```go
// GreetingListResponse response holds a list of details
type GreetingListResponse struct {
	// The total number of items
	Total uint `json:"total" example:"10"`

	// List of detail objects
	Data []*models.HelloResponse `json:"data"`
}

// Greet sends the greeting
// @Summary Greet sends the greeting
// @Description Greet sends the greeting
// @ID greet
// @Accept  json
// @Produce  json
// @Param HelloRequest body models.HelloRequest true "Hello request"
// @Success 200 {object} main.GreetingListResponse
// @Failure 400 {object} echo.HTTPError
// @Failure 404 {object} echo.HTTPError
// @Failure 500 {object} echo.HTTPError
// @Router /greet [post]
func Greet(c echo.Context) error {
  // ...
}
```

And `models.HelloResponse` is defined in a `models` pacakge within the project.

We get the following:

<img width="771" alt="image" src="https://user-images.githubusercontent.com/2081100/46247336-30959480-c3e0-11e8-9483-61656a2c9897.png">

<img width="865" alt="image" src="https://user-images.githubusercontent.com/2081100/46247340-38edcf80-c3e0-11e8-8d7a-da6baf773ef5.png">

**To Reproduce**
Sample source code can be found in [embedded_struct branch of this repo](https://github.com/bojand/swag-test/tree/embedded_struct).

**Expected behavior**
The documentation should be resolved properly.

**Your swag version**
e.g. v1.3.2

OS: MacOS High Sierra 10.13.6

Not sure if this is related or same issue as #218, #216, or #183.

dmarchin:
Also having this issue. 

My struct:

```
type data struct {
	Files    []*models.ModelsFileEntry  `json:"files"`
}
```

The swagger file generated doesn't include the correct type:
```
"files": {
    "type": "array",
    "items": {
        "type": "\u0026{%!s(token.Pos=926) %!s(*ast.SelectorExpr=\u0026{0xc42046ee60 0xc42046ee80})}"
    }
}
```

moskal91:
I got similar error, but when using model from external package outside the project. If model is nested, for example:

```
type A struct {
   Nest   B
}

type B struct {
   Value string
}
```

I got result like:
```yaml
definitions:
  extpkg.A
    properties:
      nest:
        type: B
```

Type B is not resolved. Locally within project it works ok.

[...]

PR #244, opened 2018-11-10
Add support for map and struct pointer types and selector in array pointer

bojand:
Add support for map and struct pointer types. 
Add support for selector expression in array pointer type 
Should address #221.


easonlin404:
@bojand could you help fix testing against `go1.11`?

bojand:
Hmm weird I cannot reproduce on macOS with go 1.11.2. Will investigate further.

easonlin404:
CI have passed after re-trigger job.

Issue #276, opened 2019-01-02
Multiline description does not work


kolaente:
**Describe the bug**
With #219, we got multiline descriptions. However, these don't work.

**To Reproduce**
1. Add a multiline description like so:
```
// @description Lorem Ipsum
// @description # Authorization
// @description <!-- ReDoc-Inject: <security-definitions> -->
```

**What I got**
```json
```
"info": {
        "description": "\u003c!-- ReDoc-Inject: \u003csecurity-definitions\u003e --\u003e",
        "title": "...",
...
```

**Expected behavior**
```json
```
"info": {
        "description": "Lorem Ipsum <br/># Authorization</br>\u003c!-- ReDoc-Inject: \u003csecurity-definitions\u003e --\u003e",
        "title": "...",
...
```

**Your swag version**
v1.4.0

kolaente:
Looks like it wasn't implemented for general api info, I've created a pr:  #278

PR #278, opened 2019-01-03
Multiline description for api info

kolaente:
**Describe the PR**
This adds multiline description for the general api info. It also changes the behaviour of the multiline description in the routes definition to use `\n` instead of `<br/>` as this is generally considered bad practice.

**Relation issue**
#276

### DIFF
--- a/README.md
+++ b/README.md
@@ -472,6 +472,16 @@ Field Name | Type | Description
 
 ## TIPS
 
+### Descriptions over multiple lines
+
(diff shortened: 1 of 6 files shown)


kolaente:
Thanks for merging this so fast!

lfaoro:
@kolaente how do we use this since it's not released? Is it go gettable?

kolaente:
@lfaoro The last release was at 18.2., this was merged the 22 jan so it is in the release.

Issue #266, opened 2018-12-14
Swaggertype Override Bug


Jed-Giblin:
**Describe the bug**
When attempting to override the type of a field using the swaggertype tag, there is extra parsing logic that occurs due to the ability to specify an array of some type. This logic is slightly flawed and prevents the use of a more basic type of override.

E.g. I have a custom type "JsonTime" that I want swagger to represent as a "String" but I don't want to make that type an alias for the golang String type.

**To Reproduce**

Using the below types
```
type JsonTime struct {
	time.Time
}

type Set struct {
	ReleaseDate    JsonTime 	`json:"release_date" db:"release_date" swaggertype:"string"`
}
```
Attempting to run `swag init` will produce an error + stack trace:
```
panic: JsonTime is not basic types

goroutine 1 [running]:
github.com/swaggo/swag.CheckSchemaType(0xc000337bb0, 0x8)
        /Users/jed.giblin/go/src/github.com/swaggo/swag/schema.go:8 +0x213
```

**Expected behavior**
I expected to be able to override the swagger generated type of a custom struct

**Your swag version**
master

**Desktop (please complete the following information):**
- OS [17.7.0 Darwin Kernel Version 17.7.0]

**Additional context**
I was able to prep and test a small fix, but without a PR at present.

Changes to parser.go starting at line 815
```
	if typeTag := structTag.Get("swaggertype"); typeTag != "" {
		parts := strings.Split(typeTag, ",")
                // If there is more then one item in the slice, the type is an array of a basic type
		if len(parts) == 2 {
			newSchemaType := parts[0]
			newArrayType := structField.arrayType
			if len(parts) >= 2 && newSchemaType == "array" {
				newArrayType = parts[1]
			}

			CheckSchemaType(newSchemaType)
			CheckSchemaType(newArrayType)
			structField.schemaType = newSchemaType
			structField.arrayType = newArrayType
		} else {   // Otherwise, the user simply attempted to override the type determination
			newSchemaType := parts[0]
			CheckSchemaType(newSchemaType)
			structField.schemaType = newSchemaType
		}
	}
```

easonlin404:
fix via #290

PR #290, opened 2019-01-23
add: surpport override swaggertype tag especially for self-defined st…

sdghchj:
I've run into the same problem as: #266 
This pull request gives a  solution.


easonlin404:
@sdghchj can you help add tests?

sdghchj:
@easonlin404 added

PR #295, opened 2019-02-01
add extension info to struct field

lutaoact:


Nerzal:
Could you add a test method for this feature?

lutaoact:
@Nerzal I have editted the test files to cover my code.

README.md
```diff
@@ -572,5 +572,30 @@ type Account struct {
...
+        "id": {
+            "type": "string",
+            "x-nullable": true,
+            "x-abc": "def"
+        }
+    }
+}
+```
+
+Maybe some other tools will use extension info. For example, [go-swagger/go-swagger](https://github.com/go-swagger/go-swagger) use `x-nullable` to render struct field as pointer when generate client. [reference](https://goswagger.io/use/models/schemas.html)
```
easonlin404: @lutaoact can you delete this paragraph? 

> Maybe some other tools will use extension info. For example, [go-swagger/go-swagger]

lutaoact: @easonlin404 ok, done.

PR #326, opened 2019-03-19
use go mod

pei0804:
**Describe the PR**
If use `go mod` We have to change to go 1.11 a minimum version. What do you think?
I thought contributor want to use same dependency so I wanto to quit use `go get`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,20 +2,12 @@ language: go
 sudo: false
 
 go:
-  - 1.9.x
-  - 1.10.x
   - 1.11.x
-
-matrix:
-  fast_finish: true
-  include:
-  - go: 1.11.x
-    env: GO111MODULE=on
+  - 1.12.x
 
 install:
-  - make install
-  - if [[ "${GO111MODULE}" = "on" ]]; then make tools; fi
-  
+  - make download
+
 script:
   - make fmt-check
   - make lint
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/modern-go/reflect2 v1.0.1 // indirect
 	github.com/pkg/errors v0.8.1
 	github.com/satori/go.uuid v1.2.0
-	github.com/stretchr/testify v1.2.2
+	github.com/stretchr/testify v1.3.0
 	github.com/swaggo/gin-swagger v1.0.0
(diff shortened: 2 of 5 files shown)


go.mod
```diff
@@ -13,7 +13,7 @@ require (
 	github.com/modern-go/reflect2 v1.0.1 // indirect
 	github.com/pkg/errors v0.8.1
 	github.com/satori/go.uuid v1.2.0
-	github.com/stretchr/testify v1.2.2
+	github.com/stretchr/testify v1.3.0
```
pei0804: `1.2.2` have some problem
https://travis-ci.org/swaggo/swag/jobs/508357216#L662

pei0804:
@easonlin404 review plz.

.travis.yml
```diff
@@ -2,20 +2,12 @@ language: go
 sudo: false
 
 go:
-  - 1.9.x
-  - 1.10.x
```
easonlin404: I think that shouldn't delete against 1.9/1.0 testing because users might still be use older go version.

pei0804:
I have one question.
1.9 and 1.10 doesn't have go mod.
Does They use `go get` ?
@easonlin404

.travis.yml
pei0804: Question for `use older go version.`
Old go can't use swag binary? or You say use case is contributor write swag code?

easonlin404: user still can work if they still use go 1.9/1.10 version.

easonlin404:
> I have one question.
> 1.9 and 1.10 doesn't have go mod.
> Does They use `go get` ?
> @easonlin404

users only can get package via using `go get` when go 1.9/1.10.

.travis.yml
pei0804: I understand.

pei0804:
> users only can get package via using go get when go 1.9/1.10.

I wanted to match recognition for you.
Thank you. I create another PR.

BTW. User required use mod from 1.12x?

easonlin404:
> > users only can get package via using go get when go 1.9/1.10.
> 
> I wanted to match recognition for you.
> Thank you. I create another PR.
> 
> BTW. User required use mod from 1.11x?

sure, you can reference https://github.com/golang/go/wiki/Modules docs.

PR #336, opened 2019-03-21
fix govet error

pei0804:
**Describe the PR**
fix for govet error and add go vet check.


Makefile
```diff
@@ -55,6 +55,7 @@ lint:
 	fi
 	
 	for PKG in $(PACKAGES); do golint -set_exit_status $$PKG || exit 1; done;
+	go vet
```
easonlin404: maybe can add new .PHONY:vet like:

``` sh
.PHONY: vet
vet:
 $(GO) vet $(PACKAGES)
```

easonlin404: and add to https://github.com/swaggo/swag/blob/889705a8db7763121543be8c314d836ebf040d42/.travis.yml#L19-L23

pei0804: ok

pei0804:
@easonlin404 review plz

PR #337, opened 2019-03-21
add check go vet in lint target.

pei0804:
**Describe the PR**
add go vet check.


pei0804:
with.
https://github.com/swaggo/swag/pull/336

Issue #340, opened 2019-03-22
Operation description gets appended to general info if they share the same file


umanovskis:
**Summary**

Swag version 1.4.1.

If general API info is in the same file as operations, then the @description tag of operations will also get appended to the general info. I think this was introduced after adding support for multiline @description in the general API.

**Example**

Code in the same file:

```
// @title       Test API
// @version     0.1
// @description General API description

// @Summary Returns true
// @Description Always returns true, really
// @Success 200
// @Router /test [get]
func op() bool {
        return true
}
```
After `swag init`, it will generate this description for the general API:

```
"description": "General API description\nAlways returns true, really",
```

And the operation's description itself will be correct:

```
 "description": "Always returns true, really",
```

pei0804:
Thank you!
I will review your code.

umanovskis:
PR has been merged - thanks to the maintainers.

PR #341, opened 2019-03-22
Fix handling of multiline description blocks

umanovskis:
Corrects a bug when mixing general API and ops with @description tags in the same file.

See issue #340


pei0804:
LGTM

parser.go
```diff
@@ -135,7 +140,7 @@ func (parser *Parser) ParseGeneralAPIInfo(mainAPIFile string) error {
 				case "@description":
 					if parser.swagger.Info.Description == "{{.Description}}" {
 						parser.swagger.Info.Description = strings.TrimSpace(commentLine[len(attribute):])
-					} else {
+					} else if multilineBlock {
```
pei0804: I think Below code insted of `multiLineBlock` more simple.
What do you think?
```
					if parser.swagger.Info.Description == "{{.Description}}" {
						parser.swagger.Info.Description = strings.TrimSpace(commentLine[len(attribute):])
					} else if previousAttribute == attribute {
						parser.swagger.Info.Description += "\n" + strings.TrimSpace(commentLine[len(attribute):])
					}
```

umanovskis: It's a bit simpler, but I think it creates a higher chance of reintroducing the bug later with another tag. I wanted `multilineBlock` specifically because it moves the `previousAttribute == attribute` to the previous scope before the `switch`, and then anyone changing more tags in the future will notice it.

Up to you, I can simplify per your comment if you still think that's better.

pei0804: I see. I think good your idea.

Issue #348, opened 2019-03-23
error use insted of panic


pei0804:
swag doesn't use panic.

```
❯ git grep "panic" | grep -v "test"
operation.go:                   panic(fmt.Errorf("enum value %s can't convert to %s err: %s", value, schemaType, err))
operation.go:                   panic(fmt.Errorf("enum value %s can't convert to %s err: %s", value, schemaType, err))
operation.go:                   panic(fmt.Errorf("enum value %s can't convert to %s err: %s", value, schemaType, err))
operation.go:           panic(fmt.Errorf("%s is unsupported type in enum value", schemaType))
parser.go:              panic("@scope is empty")
parser.go:                              panic("@scope can't use comma(,) get=" + v)
parser.go:              panic(fmt.Errorf("can't parse numeric value of %q tag: %v", tagName, err))
parser.go:              panic(fmt.Errorf("can't parse numeric value of %q tag: %v", tagName, err))
parser.go:                      panic(fmt.Errorf("example value %s can't convert to %s err: %s", exampleValue, schemaType, err))
parser.go:                      panic(fmt.Errorf("example value %s can't convert to %s err: %s", exampleValue, schemaType, err))
parser.go:                      panic(fmt.Errorf("example value %s can't convert to %s err: %s", exampleValue, schemaType, err))
parser.go:              panic(fmt.Errorf("%s is unsupported type in example value", schemaType))
property.go:// getPropertyName returns the string value for the given field if it exists, otherwise it panics.
property.go:    panic("not supported" + fmt.Sprint(field.Type))
schema.go:// CheckSchemaType begins panicking if typeName is not a name of primitive type
schema.go:              panic(fmt.Errorf("%s is not basic types", typeName))
swagger.go:             panic("swagger is nil")
swagger.go:             panic("Register called twice for swag: " + name)

```

PR #350, opened 2019-03-23
refactor: return error instead of panic

easonlin404:
#348


pei0804:
parser have one panic.

https://github.com/swaggo/swag/blob/91a2126fb1222a0a0493610bff9790925d631fe1/parser.go#L344-L354

PR #353, opened 2019-03-24
use trim spaced string

pei0804:
**Describe the PR**
use trim spaced string.


pei0804:
@easonlin404 review plz.

PR #355, opened 2019-03-25
refactor(parser): using error instead of panic

easonlin404:
Having exists duplicate code snippets, will refactor better with other new PR.
ref #348


easonlin404:
Having exists duplicate code snippets, I will refactor better with other new PR.

PR #362, opened 2019-03-27
[WIP] chore(travis): add fossa webhooks

easonlin404:


easonlin404:
It seems like doesn't work to notify fossa, please don't  merged.

PR #383, opened 2019-04-29
doc(readme): fix Param Type option desc

Penglq:
Looking at it is a bit confusing。Is that so?


easonlin404:
yes, thanks for your contributions.

PR #399, opened 2019-05-19
refactor: return error instead of panic

easonlin404:
fix https://github.com/swaggo/swag/issues/348


easonlin404:
we can close #348 after merged it.

pei0804:
Nice

Issue #402, opened 2019-05-31
Panic when use `type DateOnly time.Time` as struct field


Kipriz:
**Describe the bug**
Swag cmd fails with panic when I use the following type alias:
##### api/api.go
```
// @Success 200 {object} data.TimeContainer	"ok"
func GetTimeAsTimeContainer(c *gin.Context) {
}
```
##### data/alias.go
```
type TimeContainer struct {
	Name      types.StringAlias `json:"name"`
	Timestamp time.Time         `json:"timestamp"`
	CreatedAt types.DateOnly    `json:"created_at"`
}
```
##### types/alias.go
```
type StringAlias string
type DateOnly time.Time
```

When `DateOnly` is declared in the same file as `TimeContainer`, then `swaggertype:"primitive,string"` prevents the panic. However, this annotation does not work when `DateOnly` is declared in another package.

**Expected behavior**
Swag should output:
```
   "definitions": {
        "data.TimeContainer": {
            "type": "object",
            "properties": {
                "created_at": {
                    "type": "string"
                },
                "name": {
                    "type": "string"
                },
                "timestamp": {
                    "type": "string"
                }
            }
        }
    }
```

**Your swag version**
e.g. 1.5.0

**Your go version**
e.g. 1.12.1

**Desktop (please complete the following information):**
 - OS: Windows 10

**Additional context**

easonlin404:
fixed by #403

PR #403, opened 2019-05-31
Fix processing alias type in struct when alias type is in another package

Kipriz:
**Describe the PR**
Add ability to use alias types from another package.

**Relation issue**
https://github.com/swaggo/swag/issues/402


easonlin404:
@Kipriz Thanks for your contributions.

PR #413, opened 2019-06-18
Errormessage does now include tagname

Nerzal:
**Describe the PR**
Improved error message, when a markdown file is missing. It does now include the tag name.


parser.go
```diff
@@ -423,7 +423,7 @@ func getMarkdownFileForTag(tagName string, dirPath string) (string, error) {
 		}
 	}
 
-	return "", errors.New("Unable to find Markdown file in the given directory")
+	return "", errors.New("Unable to find Markdown file for tag '" + tagName + "' in the given directory")
```
easonlin404: use `fmt.Errorf(format string, a ...interface{})` more suitable if needs formats string.

Nerzal: ok

ubogdan:
@Nerzal it's a while sicen you started this PR , so in order to fix this I'm going to another PR for this.

Nerzal:
thx @ubogdan

Issue #422, opened 2019-06-28
form is not supported paramType


Cloudzp:
**Describe the bug**
use command ` swag.exe init --parseVendor=true` when pkg "github.com/swaggo/swag" in my vendor, i get an error :" ParseComment error in file vendor\github.com\swaggo\swag\operation.go :form is not supported paramType "

**Expected behavior**
get a success result, with `--parseVendor=true`

**Screenshots**
![image](https://user-images.githubusercontent.com/25190767/60312275-5494ff00-998d-11e9-8cff-37aba3a96e9a.png)

**Your swag version**
$ swag.exe -v
swag.exe version v1.5.1

**Your go version**
$ go version
go version go1.11 windows/amd64

easonlin404:
fixes via #423

ismdeep:
@easonlin404 this bug remained at v1.8.3

PR #423, opened 2019-06-28
chore: fix example comment at ParseParamComment func

Cloudzp:
**Describe the PR**
when use `swag init --parserVendor=true ` with pkg "github.com\swaggo\swag" in you vendor, will get an error:  ParseComment error in file E:\workspace\src\park_record\vendor\github.com\swaggo\swag\operation.go :form is not supported paramType

**Relation issue**
#422


operation.go
```diff
@@ -118,7 +118,7 @@ func (operation *Operation) ParseComment(comment string, astFile *ast.File) erro
 var paramPattern = regexp.MustCompile(`(\S+)[\s]+([\w]+)[\s]+([\S.]+)[\s]+([\w]+)[\s]+"([^"]+)"`)
 
 // ParseParamComment parses params return []string of param properties
-// E.g. @Param	queryText		form	      string	  true		        "The email for login"
+// E.g. @Param	queryText		formData	      string	  true		        "The email for login"
```
easonlin404: why this change can address #422 ?

Cloudzp: paramType 'form' is not supported, but in you sample give a paramType "form", this can result `swag init` failed with error "paramType form not supported"

easonlin404:
@Cloudzp Thanks for your contributions.

PR #451, opened 2019-07-13
parse object description from structure annotation.

zhy1:
**Describe the PR**
parse object description from structure annotation. 

some entity.go
```go

type Post struct {
	ID int `json:"id" example:"1" format:"int64" description:"Post primary field."`
	// Post name
	Name string `json:"name" example:"poti"`
	// Post data
	Data struct {
		// Post tag
		Tag []string `json:"name"`
	} `json:"data"`
}
```

output
swagger.json
```json
 "id": {
                    "description": "Post primary field.",
                    "type": "integer",
                    "format": "int64",
                    "example": 1
                },
```

**Additional context**
if attribute not setting. use the previous rules


operation.go
```diff
@@ -220,6 +220,10 @@ var regexAttributes = map[string]*regexp.Regexp{
 	"maxlength": regexp.MustCompile(`(?i)maxlength\(.*\)`),
 	// for format(email)
 	"format": regexp.MustCompile(`(?i)format\(.*\)`),
+	// for format(email)
```
easonlin404: comment should be for `description`.

operation.go
```diff
@@ -220,6 +220,10 @@ var regexAttributes = map[string]*regexp.Regexp{
...
+	"description": regexp.MustCompile(`(?i)description\(.*\)`),
+	// for format(email)
```
easonlin404: as above

Issue #466, opened 2019-07-19
Backticks break spec generation


zaynetro:
**Describe the bug**

My use case: Generate `swagger.{yaml/json}` from code comments.

Bug: Having backticks in e.g description field break consequtive spec generations.

**To Reproduce**

Define spec:

```
// @title Gobpg API
// @version 1.0
// @description Something `wrapped`.
```

Generate spec:

```
# Init swagger spec for the first time
$ swag init
...
2019/07/19 12:40:45 create docs.go at  docs/docs.go
2019/07/19 12:40:45 create swagger.json at  docs/swagger.json
2019/07/19 12:40:45 create swagger.yaml at  docs/swagger.yaml

# While having existing `docs/` directory spec generation fails
$ swag init -g gobpg.go
2019/07/19 12:41:17 Generate swagger docs....
2019/07/19 12:41:17 Generate general API Info, search dir:./
2019/07/19 12:41:17 ParseFile error:docs/docs.go:19:98: expected ';', found '{' (and 2 more errors)
```

Backticks break string definition in `docs.go`.

**Expected behavior**

Files are generated every time `swag init` is run. 

This bug doesn't prevent me from using swag. I can just remove `docs` directory or `docs/docs.go` file every time before generating the spec.

Options that would solve my issue:

1. Ignore parsing `docs.go` file errors when generating the spec
1. Add an option to skip generating `docs.go` file completely
1. Fix backticks in strings breaking Go syntax

**Your swag version**: latest master

**Your go version**: 1.12.6

**Desktop (please complete the following information):**
 - OS: macOS Mojave

ubogdan:
@zaynetro  please use 
```
$ go get -u github.com/swaggo/swag/cmd/swag@master and try again.
$ rm docs/docs.go
$ swag init -g gobpg.go
```

zaynetro:
Do you recommend to keep removing `docs/docs.go` before every spec generation as a solution?

As mentioned in the issue this does solve the problem for me but that doesn't solve the underlying issue in the project.

If don't feel that this issue needs fixing then feel free to close it.

ubogdan:
sorry , i made a confusion. We fixed Backticks in struct comments. I will do a pr for main api generation soon.

zaynetro:
No worries. Could it be done that do not remove backticks but instead fix Go syntax by e.g. string contatenation?

I am open to contribute as well if you tell me what the fix should be.

ubogdan:
[...]

PR #468, opened 2019-07-19
refactor(gen.go): improve code coverage

ubogdan:
**Describe the PR**
Fix issue #466
Fix issue #466
Refactor gen/gen.go , improve gen.go coverage

**Relation issue**
#466, #466

**Additional context**
None.


gen/gen.go
```diff
@@ -110,32 +108,115 @@ func (g *Gen) Build(config *Config) error {
...
 }
 
-var packageTemplate = template.Must(template.New("").Parse(`// GENERATED BY THE COMMAND ABOVE; DO NOT EDIT
+func (g *Gen) jsonIndent(data interface{}) ([]byte, error) {
+	return json.MarshalIndent(data, "", "    ")
+}
+
+func (g *Gen) formatSource(src []byte) []byte {
+	code, err := format.Source(src)
+	if err != nil {
```
easonlin404: why does return code instead of return `errr`?

ubogdan: If formating goes wrong , mod.go is empty. You won't be able to see what is wrong. 
If you want other behavior please let me know and i will change it that way.

Issue #357, opened 2019-03-26
string raw in body


dimuls:
**Describe the bug**
If you try to make body string param you will get wrong generated docs:
```
<...>
      - description: This is body string param
        in: body
        name: body-string-param
        required: true
        schema:
          $ref: '#/definitions/string'
          type: object
<...>
```

**To Reproduce**
Use this kind of definition:
```
// @param body-string-param body string true "This is body string param"
```

**Expected behavior**
I'm not sure in right definition in docs. But certainly this is wrong definition:
```
schema:
          $ref: '#/definitions/string'
          type: object
```
It causes error in gin-swagger ui.

**Your swag version**
1.4.1

**Your go version**
1.11.2

pei0804:
I think you have to use struct.
`// @param body-string-param body string true "This is body string param"`

https://github.com/swaggo/swag/blob/master/testdata/simple/api/api.go#L13

dimuls:
No need string exactly. I have `@accept text/csv` API method. I need to fill textarea with CSV rows in UI.

pei0804:
Is this POST method?

dimuls:
Yes, this is POST method.

pei0804:
Does this api no need key of key-value?
Is this similar api? https://stackoverflow.com/questions/50998620/post-csv-text-file-using-curl

BTW, swag have file upload annotation.
eg: https://github.com/swaggo/swag/blob/master/testdata/simple/api/api.go#L50

dimuls:
My API does not require any key. CSV data will be read from whole body.

pei0804:
swag don't support this case yet.
Does @easonlin404  know?

Maybe this.
https://swagger.io/docs/specification/2-0/describing-request-body/


```yaml
    paths:
      /status:
        post:
          summary: Updates the status message.
          consumes:
            - text/plain      # <----------
        parameters:
          - in: body
            name: status
            required: true
            schema:
              type: string  # <----------
        responses:
          200:
            description: Success!
```

pei0804:
> About file upload: can i do something like this?

Yes.

`// @Param  filekey formData file true  "this is a test file"`

dimuls:
> 
> 
> swag doesn't support this case yet.
> Does @easonlin404 know?
> 
> Maybe this.
> https://swagger.io/docs/specification/2-0/describing-request-body/
> 
> ```yaml
>     paths:
>       /status:
>         post:
>           summary: Updates the status message.
>           consumes:
[...]

Issue #299, opened 2019-02-12
 ParseDefinitions not supported 'Array' yet.


2BAB:
**Describe the bug**
I saw there is `{array}` definition in response annotation, but I don't know how to make an array as the parameter. Here is the example:

`// @Param  notification  body  push.ModelList  true  "balabala"`

`type ModelList []Model`

Then I got: `panic: ParseDefinitions not supported 'Array' yet.`

**To Reproduce**
1. Run `swag init`

**Expected behavior**
Support array in parameters annotation.

**Your swag version**
github.com/swaggo/gin-swagger v1.0.0
github.com/swaggo/swag v1.4.0

xiaoguiSniper:
```
type X struct {
  IDs []int64 
}
```
`// @Param Message body api.X true "tips"`
when I use `swag init`,I got this message in api difinition:`Unknown Type:int64`
so,I wonder if any solution for this question? 
or, this is just my fault for `body` type which I want to define an array type.

ubogdan:
Issue #299 supporting array of things is already fixed on #462.
Annotation
`// @Param payload body web.TagList true "description"
`


```
package web

type TagList []Tag

// Tag example
type Tag struct {
	ID   int    `json:"id"`
	Name string `json:"name"`
}
```

Please get latest swagger and open an issue if still not working.

Issue #83, opened 2018-03-29
Add Array and Multi-Value Parameters


pei0804:
https://swagger.io/docs/specification/2-0/describing-parameters/

```yaml
      # color=red,black,white
      parameters:
        - in: query
          name: color
          type: array
          collectionFormat: csv
          items:
            type: string
```

```
// @Param some_id []query true "Some ID" uniqueItems(true) minnux(0)
```

pei0804:
```go
// @Param some_id []query true "Some ID"
```

easonlin404:
PR is welcome! 👍

AyrtonRicardo:
@easonlin404 @pei0804 I've made a PR with this change, maybe we can merge it :D I'm also looking for this feature.

ubogdan:
[]query is confusing , "query", "body" describes where the parameter should be located. 
```
// @Param some_id []query true "Some ID" uniqueItems(true) minnux(0)
```

PR #469, opened 2019-07-20
Add possibility to describe a body request as array

AyrtonRicardo:
**Describe the PR**
This PR was made because I faced the same issue as #299 and #83 
I was also trying to add more formats to the body like describe in #357 but I saw @ubogdan was working on it, so I just fixed my problem with array body. 

**Relation issue**
- [ParseDefinitions not supported 'Array' yet](https://github.com/swaggo/swag/issues/299)
- [Add Array and Multi-Value Parameters](https://github.com/swaggo/swag/issues/83)


ubogdan:
Issue #299 supporting array of things is allready implemented.
Annotation
`// @Param payload body web.TagList true "description"
`


```
package web

type TagList []Tag

// Tag example
type Tag struct {
	ID   int    `json:"id"`
	Name string `json:"name"`
}
```

testdata/array_body/api/api.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"github.com/gin-gonic/gin"
+)
+
+// PostPosts godoc
+// @Summary Add a lot of posts to somewhere
+// @Description Post lot of posts to somewhere
+// @Accept  json
+// @Produce  json
+// @Param posts body []web.Post true "This is a lots of posts"
```
ubogdan: Why can't it be 
`// @Param posts body web.PostCollection true "This is a lots of posts"
`
and web.PostCollection defined as ?
`type PostCollection []Post
`

operation.go
```diff
@@ -155,15 +155,36 @@ func (operation *Operation) ParseParamComment(commentLine string, astFile *ast.F
 
 	//five possible parameter types.
 	switch paramType {
+	case "[]query":
```
ubogdan: So "query" means the parameter should be sent over query. 
Now what "[]query" is ?

operation.go
```diff
@@ -155,15 +155,36 @@ func (operation *Operation) ParseParamComment(commentLine string, astFile *ast.F
...
+			return fmt.Errorf("%s is not supported type for parameter %s", validType, paramType)
+		}
+		param = createParameter(paramType, description, name, validType, required)
 	case "query", "path", "header":
 		param = createParameter(paramType, description, name, TransToValidSchemeType(schemaType), required)
 	case "body":
-		param = createParameter(paramType, description, name, "object", required) // TODO: if Parameter types can be objects, but also primitives and arrays
-		if err := operation.registerSchemaType(schemaType, astFile); err != nil {
+		objType := "object"
+		if strings.HasPrefix(schemaType, "[]") {
```
ubogdan: I know "it works" but this is not an elegant way.
You should do the job in  registerSchemaType because you have access to ast and you can identify what ever it is.

parser_test.go
```diff
@@ -2766,3 +2766,166 @@ func TestParseOutsideDependencies(t *testing.T) {
 		t.Error("Failed to parse api: " + err.Error())
 	}
 }
+
+func TestArrayBody(t *testing.T) {
```
ubogdan: This test is useless. We are testing the same thing here:
https://github.com/swaggo/swag/blob/c19280e76fac72a09e9730a4b3c9ae3e144d9b3a/parser_test.go#L2246-L2303

ubogdan:
Since you didn't go forward with this i pushed #489 .

AyrtonRicardo:
@ubogdan thanks for merge it, I was in vacation without internet. :bowtie:

Issue #474, opened 2019-07-23
error resolving host and basepath


bruceadowns:
**Describe the bug**
The latest version (tips) of swag is not resolving host and base path.

i.e.
`curl -X GET "https://{{.Host}}{{.BasePath}}/api/version" -H  "accept: text/plain"`

**To Reproduce**
1. generate docs.go
2. compile
3. host swagger
4. See invalid url

**Expected behavior**
This:
`curl -X GET "https://{{.Host}}{{.BasePath}}/api/version" -H  "accept: text/plain"`

should be:
`curl -X GET "https://localhost:8443/api/version" -H  "accept: text/plain"`

**Screenshots**
![image](https://user-images.githubusercontent.com/1503908/61683082-01cb1080-acc9-11e9-9cf3-e71b17e9bed3.png)

**Your swag version**
~~1.5.1 - tips~~
tips

**Your go version**
1.12.7 - latest

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: firefox
 - Version: 68

**Additional context**

ubogdan:
Please update to [latest](https://github.com/swaggo/swag/releases) version (1.6.2) and let us know if this still happens.

bruceadowns:
To clarify, v1.5.1 and v1.6.2 work fine. The problem occurs when using tips.

easonlin404:
@bruceadowns How do you show swagger doc site? I think you directly use `swagger.json `or `swagger.ymal` generated by swag cliright? 

@ubogdan it's kind of a bug caused by programmatically setting values because swag output json/yaml but doesn't pass programmatically value to it.

bruceadowns:
I first get swag:
`go get github.com/swaggo/swag/cmd/swag`

then init:
`swag init -g cmd/cnc.go`

then route and host:
`router.PathPrefix("/swagger").Handler(httpSwagger.WrapHandler)`

ubogdan:
@bruceadowns i have no clue how to download and test "tips" version . 
If you speak about lastest commit , I can certify  "it works on my computer" .
```
package main

import (
	"github.com/gorilla/mux"
	httpSwagger "github.com/swaggo/http-swagger"
	"github.com/swaggo/swag/example/is461/docs"
	"net/http"
)

// @title GEO service
// @version 1.0
// @description Returns GEO information from a request.
// @description Something `wrapped`.
// @host localhost:1323
// @schemes http https
// @BasePath /

func main() {
	router := mux.NewRouter()
[...]

Issue #461, opened 2019-07-18
Descriptions for path interfere with info description


pjebs:
See **both** `<------------- SEE HERE` below:

```

// @title GEO service
// @version 1.0
// @description Returns GEO information from a request.  <------------- SEE HERE

// @host localhost:1323
// @BasePath /

func main() {
	http.HandleFunc("/healthz", HealthHandler)
	http.HandleFunc("/geo", GeoHandler)
}

// @Summary Return geo info
// @Description Retrieves geo related info.
// @Description If an ip address is not provided, the IP address of the incoming request is used.   <------------- SEE HERE
// @Tags geo
// @Produce  json
// @Param ip_address query string false "IP Address"
// @Success 200 {object} main.GeoResponse
// @Failure 500 {string} string "TODO "
// @Router /geo [get]
func GeoHandler(w http.ResponseWriter, req *http.Request) {
	defer func() { uerrors.CatchError(recover(), nil) }()

```

pjebs:
![image](https://user-images.githubusercontent.com/5377970/61435119-22b6fd00-a97b-11e9-8258-9dd49738e4e7.png)

You can see the description from the route leaking into the info description.

ubogdan:
This is happening due multi line comment and the fact you declared the http.handler functions in the same file General API annotation.

ubogdan:
there is no bug with ast, you just not following the templates and swagger is not able to parse data correctly.

PR #475, opened 2019-07-25
fix: error resolving host and basepath

ubogdan:
**Describe the PR**
fix issue 474
fix issue 461
repair case tests

 
**Relation issue**
#474 
#461 

**Additional context**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 dist
 testdata/simple*/docs
+example/basic/docs/*
+example/celler/docs/*
 cover.out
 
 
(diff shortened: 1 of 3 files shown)


easonlin404:
LGTM, but this change I'm not sure does it work for programmatically value passing? I think we could add the test for it.

ubogdan:
This change fix annoying {{.Host}} and {{.BasePath}} into json/yaml files.

ubogdan:
I've left some comments on issue itself. Anyway if you don't like it just refuse the PR.

easonlin404:
It’s ok for me. I would check later and add the tests if necessary.

Issue #478, opened 2019-07-29
Use template variables like `Schemes` and `Host` in the API Description


lonnblad:
**Is your feature request related to a problem? Please describe.**
I would like to be able to use template variables like `Schemes` and `Host` in my API Description like this: `// @Description More information can be found here: [Service Documentation]({{index .Schemes 0}}://{{.Host}}/docs/service).` This is because we host other information like flow diagrams, code examples, etc. on the same server as the Open API spec.

**Describe the solution you'd like**
Parse and Execute the SwaggerInfo.Description as a template like the generated doc.

ubogdan:
According to your example please explain what will happen if host and schemes will not be set in comments/annotation (because they are optional). 
Is there any reason this doesn't work for you ?
```
// @Description More information can be found here: [Service Documentation](/docs/service).
```

lonnblad:
true, that should cover my case

lonnblad:
close the issue and revisit if another use-case comes up?

ubogdan:
I think is better to not implement it for the moment. If we will have more requests for this we will take in consideration this feature.
Thanks.

PR #479, opened 2019-07-29
Applies SwaggerInfo on SwaggerInfo.Description

lonnblad:
**Describe the PR**
Applies SwaggerInfo on SwaggerInfo.Description as a pre-step before applying it one the generated doc.

**Relation issue**
#478 

**Additional context**

### DIFF
--- a/gen/gen.go
+++ b/gen/gen.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"github.com/go-openapi/spec"
 	"go/format"
 	"io"
 	"log"
@@ -15,7 +14,9 @@ import (
 	"time"
 
 	"github.com/ghodss/yaml"
+	"github.com/go-openapi/spec"
 	"github.com/pkg/errors"
+
 	"github.com/swaggo/swag"
 )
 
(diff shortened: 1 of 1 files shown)


gen/gen.go
```diff
@@ -254,19 +255,25 @@ var SwaggerInfo = swaggerInfo{
...
 	t, err := template.New("swagger_info").Funcs(template.FuncMap{
 		"marshal": func(v interface{}) string {
 			a, _ := json.Marshal(v)
 			return string(a)
 		},
-	}).Parse(doc)
+	}).Parse(text)
 	if err != nil {
-		return doc
+		return text
```
ubogdan: Why are u ignoring errors ?

gen/gen.go
```diff
@@ -254,19 +255,25 @@ var SwaggerInfo = swaggerInfo{
...
 	}
 
 	var tpl bytes.Buffer
-	if err := t.Execute(&tpl, SwaggerInfo); err != nil {
-		return doc
+	if err := t.Execute(&tpl, info); err != nil {
```
ubogdan: Why are u ignoring errors ?

gen/gen.go
```diff
@@ -254,19 +255,25 @@ var SwaggerInfo = swaggerInfo{
 type s struct{}
 
 func (s *s) ReadDoc() string {
+	sInfo := SwaggerInfo
+	sInfo.Description = parseAndExecuteTemplate(sInfo.Description, sInfo)
```
ubogdan: Read doc is called at run time. why do wen need  a template with func map called twice ?

lonnblad: Because the `Description` includes template variables that will not be set today.
So the first template execution is just for the Description and the second is for the full Document.

gen/gen.go
lonnblad: The errors are ignored today, that is nothing I changed.

gen/gen.go
lonnblad: The errors are ignored today, that is nothing I changed.

gen/gen.go
ubogdan: I was asking if you can't find a better way to solve this instead of calling the template 2 times.

lonnblad: right, I see two ways;
- treat the Description as one template and the full document as one template and then apply the SwaggerInfo twice, like I did it, but I am definitely with you that it feels strange.
- apply the Description on the full document first and then apply the SwaggerInfo just once, the question would be have to do it best, something like `strings.Replace(doc, "{{.Description}}", sInfo.Description, -1)` could work

ubogdan: if you do this before rendering template , i think it may be good enough . ( more eficient then parse+execute  the template 2 times). 
```
sInfo := SwaggerInfo
sInfo.Description = strings.Replace(doc, "{{.Description}}", sInfo.Description, 1)
```

PR #491, opened 2019-08-08
Improved error message, when a markdown file is missing …

### LINKED ISSUE #432: Fail when module contains no package
**Describe the bug**
When using this with go modules: 
If the module only has sub packages and no package itself

>$ tree
.
├── a
│   └── a.go
├── b
│   └── b.go
└── go.mod

then swag init will fail. 
The error message is: 
>2019/07/02 16:36:50 execute go list command, exit status 1, stdout:, stderr:can't load package: package gitlab.com/fino/uba: unknown import path "path/to/module": cannot find module providing package path/to/module

Creating a random go file that contains the package <module folder name>, fixes the issue

I think you are using `go list` on the `./` directory, which causes this

**Expected behavior**
It should work without a redundant "root" package in the module

**Screenshots**
image

**Your swag version**
e.g. 1.6.0

**Your go version**
e.g. 1.12.6

 - OS: fedora 29

ubogdan:
**Describe the PR**
Removed pkg/errors dependency from whole project tree. 
Finish PR #413
Removed an error handler that was introduced to fix #432 . error is nil all the time , the code works the same without that piece of code.

**Relation issue**
None

**Additional context**
None.


pei0804:
Good job.

pei0804:
Why does have more dependencies in go module?

ubogdan:
if you speak about "go.mod"

1. package "github.com/gin-gonic/gin" is used in example/basic/api
2. package "github.com/satori/go.uuid" is used in example/celler/model/account.go
3. package "golang.org/x/tools" is used in operation.go

pei0804:
It seems necessary.

easonlin404:
@swaggo/swaggo-core-team IMO, I think `github.com/gin-gonic/gin` and 
`github.com/satori/go.uuid` these packages don't need to add in go.mod because all of these only be used in example.  I think add go.mod in examples code itself is better.

pei0804:
Sound good.
If so, are dependencies ignored for example code?

easonlin404:
no I mean `github.com/gin-gonic/gin` and `github.com/satori/go.uuid` these should be declare   at their own example go.mod like:
```
github.com/swaggo/swag/tree/master/example/basic/go.mod
github.com/swaggo/swag/tree/master/example/celler/go.mod
```
not added in github.com/swaggo/swag/go.mod

pei0804:
OK. I understand.

pei0804:
I create PR.

pei0804:
can not multi go.mod.

```
package github.com/swaggo/swag/example/celler: unknown import path "github.com/swaggo/swag/example/celler": ambiguous import: found github.com/swaggo/swag/example/celler in multiple modules:
```

Issue #488, opened 2019-08-07
Option for removing the generation time at top of docs.go


xizhibei:
**Is your feature request related to a problem? Please describe.**
I add the `docs` dir to my git repo, and I want add the docs generation step to git pre commit hook for automation.

But every time the hook executes, the timestamp at the top of `docs.go` changes, even no changes in other parts of `docs.go`

**Describe the solution you'd like**
A cli option for removing the timestamp.

**Describe alternatives you've considered**
1. Check the `docs.go` changes, if the content is not change, then no for the timestamp too.
2. Completely remove the timestamp, since we can always get the doc modify time and author from git.

For 2, there are some other tools used for generation won't add timestamp at their generated text.

For examples:

1. https://github.com/golang/mock
2. https://github.com/go-kit/kit

**Additional context**
None.

ubogdan:
If i understand well , your git pre commit hook should force the user to not be able to commit if the latest commited docs.go file is different from the one generated  via hook. Correct ?

xizhibei:
Yes, and thanks for your reply.

I'm using the [pre-commit](https://pre-commit.com/) and it won't let me commit if the code is modified by the hooks, but it does not matter, what I am saying is the timestamp should not be added to git.

The hooks is simple:

```yaml
- repo: local
  hooks:
  - id: generate-docs
    name: Generate docs
    pass_filenames: false
    entry: swag init --generalInfo cmd/server.go
    files: \.go$
    language: system
    types: [text]
```

ubogdan:
I was thinking to a strategy like update-only , that will change the entire file if local file is different than the generated one (excluding the first 3 lines which is the header).
Unfortunately i was busy and i had no time to advance with this.

alexanderbez:
This feature is useful for my project where we have a Makefile target/CI job that checks if annotations have been changed without updating the swagger docs. It would be helpful if `swag init` had a flag to ignore/skip the timestamp.

joshk0:
Isn't this solved by `--generatedTime` and it already defaults to false?

PR #501, opened 2019-08-28
Add generated time flag

xizhibei:
**Describe the PR**
Add generation time flag, as described below in https://github.com/swaggo/swag/issues/488

**Relation issue**
https://github.com/swaggo/swag/issues/488

**Additional context**


nnordrum:
IMO, you should add a generatedTime flag with a true default. I don’t like negative options, but maybe that’s just me.

xizhibei:
Yes, I did not find the `cli.BoolTFlag` for that at first, thks

gen/gen.go
```diff
@@ -217,8 +222,8 @@ func (g *Gen) writeGoDoc(output io.Writer, swagger *spec.Swagger) error {
 }
 
 var packageTemplate = `// GENERATED BY THE COMMAND ABOVE; DO NOT EDIT
-// This file was generated by swaggo/swag at
-// {{ .Timestamp }}
+// This file was generated by swaggo/swag{{ if .GeneratedTime }} at
+// {{ .Timestamp }}{{ end }}
```
easonlin404: Already have `Timestamp` filed why need another `GeneratedTime` to generate? 😕

gen/gen_test.go
```diff
@@ -230,7 +230,7 @@ func TestGen_writeGoDoc(t *testing.T) {
 	swapTemplate := packageTemplate
 
 	packageTemplate = `{{{`
-	err := gen.writeGoDoc(nil, nil)
+	err := gen.writeGoDoc(nil, nil, nil)
```
easonlin404: could you and the new test cases having pass by configuration?

gen/gen.go
xizhibei: You mean I should change the name or remove it? rename means changing to `EnableGeneratedTime` or etc, and remove means replacing `if .GeneratedTime` 
 with `if not .Timestamp.IsZero()`

For removing, I think `if .GeneratedTime` would be more explicit.

gen/gen_test.go
xizhibei: Sure

ubogdan:
@xizhibei , any update here ?

xizhibei:
Waiting for @easonlin404 reply, but I can make the test code update first

alexanderbez:
Looking forward to having this in a release 👍

forestgagnon:
Eagerly awaiting this as well! When code generation commands are idempotent, it really helps reduce churn and noise in repositories, and the reproducibility of output makes builds more scriptable.

gen/gen.go
easonlin404: I misunderstand your point. LGTM.

easonlin404:
sorry for the late reply.  @xizhibei can you help fix conflicting files if you have free time? Rest LGTM.

ubogdan:
Thanks for contributing.

Issue #504, opened 2019-08-30
Required tag doesn't work with nested object property


myoshi1012:
I define this struct and execute swag,
```
type Owner struct {
	Id        uint64   `json:"id" validate:"required"`
	Company   *Company `json:"company" validate:"required"`
}
```
but swag doesn't generate correctly.
```
  model.Owner:
    properties:
      company:
        $ref: '#/definitions/model.Company'
        type: object
      id:
        type: integer
    required:
    - id
    type: object
```

**Expected behavior**
expect:
```
  model.Owner:
    properties:
      company:
        $ref: '#/definitions/model.Company'
        type: object
      id:
        type: integer
    required:
    - company
    - id
    type: object
```

**Your swag version**
v1.6.2

**Your go version**
1.12.5

LennyPenny:
Can this be closed now @myoshi1012 ?

easonlin404:
fixed by #541 . Please use the latest version against the master. Reopen it if you still have any problems.

PR #514, opened 2019-09-11
Fix nested struct parsing

importcjj:
**Describe the PR**
When I use swag, I found two problems;

1. Nested structs inside can not be parsed correctly.

2. When the response struct and the API function exist in the same file, the struct can not be parsed.

**Relation issue**
https://github.com/swaggo/swag/issues/504

**Additional context**


ubogdan:
Can you please add test case for this scenario ? 
Thanks for your contribution !

ubogdan:
Great job ! Thanks for your contribution !

Issue #517, opened 2019-09-13
Use map nested in struct


lvandenbrink:
**I would like to use a map in a object**
I have the following object:
```
type Bar struct {
	Field2 string
}
type FooBarMap struct {
	Field3  map[string]Bar
}
```

**Expected behavior**
I would like to result in the following swagger definition
```
FooBarMap:
    type: object
    properties:
      field3:
        type: object
        additionalProperties:
          type: object
          properties:
            field2:
              type: string

```

**Swag version**
v1.6.2

**Go version**
go version go1.12.5 linux/amd64

lvandenbrink:
I created a PR. I updated today the code to the new master. Can someone take a look at it? @ubogdan?

PR #521, opened 2019-09-17
add support for map nested in struct

lvandenbrink:
Mappings nested in structs are not generated with additional properties. If a field is of a MapType, add the field with additional properties. If needed add the key and value type to the scheme.

**Relation issue**
https://github.com/swaggo/swag/issues/517


lvandenbrink:
I added a FooBarMap struct to the api in composition test. With view-covered I see that the tests the covers the added code. Except for when the map is required or the parsing of the value fails. 

If that doesn't suffice, can you indicate which part should further tested and where to add it?

testdata/composition/expected.json
```diff
@@ -151,6 +171,22 @@
...
+                        "type": "object",
+                        "properties": {
+                            "field4": {
+                                "type": "string"
+                            }
+                        }
+                    }
+                }
+            }
+        },
```
ubogdan: Definition 
```
type FooBarMap struct {
	Field3 map[string]MapValue
}

type MapValue struct {
	Field4 string
}
```
should produce 
```  
api.FooBarMap:
    type: object
    properties:
      field3:
        type: object
        additionalProperties:
          $ref: '#/definitions/api.MapValue'

  api.MapValue:
    type: object
    properties:
      field4:
        type: string

```

ubogdan: This will register your api.MapValue to definitions
```
		itemSchema, err := parser.parseTypeExpr(pkgName, "", astTypeMap.Value)
		if err != nil {
			return properties, nil, err
		}
```

Neighter of 2 solutions.

You need to use the astTypeMap.Value to add it to object..
```
mapValueScheme :=  &spec.Schema{
	SchemaProps: spec.SchemaProps{
		Ref: spec.Ref{
			Ref: jsonreference.MustCreateRef("#/definitions/" + pkgName + "." + astTypeMap.Value),
		},
	},
}
```

I'm quite busy these days and I don't have much time to spend here. I hope i was clear enough with the last response.

lvandenbrink: You made it clear. Almost did it yourself ><

Issue #524, opened 2019-09-18
Support readOnly attribute


wingsofovnia:
It would be great to have `readOnly: true` supported as a model attribute:
```
type Car struct {
    ID string `json:"id" readOnly:"true"` // <----
    Name string `json:"name"`
}
```

ubogdan:
Looks like "readOnly" attribute it's [OpenAPI 2.0](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md) compliant. 
I will like to do PR for it but I'm quite busy in this period . 
If you really need this feature and you have some time please contribute to swag  by writing a PR.

lvandenbrink:
I would like to have the same feature. I created a pr and added the read only to one of the tests. Let me know if it suffice.

ubogdan:
Your PR was approved . Thanks for contributing.

PR #529, opened 2019-09-23
support for read only

lvandenbrink:
support for read only.

**Relation issue**
https://github.com/swaggo/swag/issues/524


ubogdan:
Thanks for your contribution.

PR #541, opened 2019-10-12
Required tag doesn't work with nested object property #504

takushi-m:
**Relation issue**
https://github.com/swaggo/swag/issues/504


easonlin404:
@takushi-m Thanks for your contributions. 💯

Issue #551, opened 2019-11-06
@Router route doesn't work with a ':' character in it


Nephylhim:
Hi!

Firstly, thank you for this project. It's helping me to create a better doc for our products! :+1:

**Describe the bug**
I have a route using a ':' character in it and it doesn't work. Furthermore, it can erase another route if the path before the ':' is the same. ex:
/objects [get]
/objects:modifier [get]
The second route will be saved as /objects and will therefore override the real /objects route.

**To Reproduce**
1. This example works to test the behavior

```go
// @id objectList
// @Summary Objects listing
// @Router /objects [get]
// @Description A listing method...
// @Tags Test
// @Param q query string false "Isn't kept either"
func objectList() {
}

// @id specialObjectList
// @Summary Special objects listing
// @Router /objects:modifier [get]
// @Description A special listing method...
// @Tags Test
func specialObjectList() {
}
```

2. swag init
3. See the result

![Screenshot_2019-11-06 Swagger UI](https://user-images.githubusercontent.com/7249952/68308374-0fa47600-00ad-11ea-8710-4f0b9e950269.png)

Here is the resulting JSON if you prefer (in the "paths" object):
```json
{
    "/objects": {
        "get": {
            "description": "A special listing method...",
            "tags": [
                "Test"
            ],
            "summary": "Special objects listing",
            "operationId": "specialObjectList"
        }
    }
}
```
The real "/objects" route is overriden by the "/objects:modifier" route but is set as "/objects".

**Expected behavior**
To have the "/objects" and the "/objects:modifier" routes correctly registered

**Your swag version**
swag version v1.6.3

**Your go version**
go version go1.10.7 linux/amd64

**Desktop (please complete the following information):**
 - OS: linux, Debian Buster

**Additional context**
I think I have found where the problem occurs:
https://github.com/swaggo/swag/blob/master/operation.go#L440

On my early search, I found that adding ':' in the regex should do the job without causing any side effect: https://regex101.com/r/t6AXRt/1
I plan to try to fix/test it along with a MR.
Is `operation_test.go` the only place where you'd want a unit test?

EDIT: the ":modifier" is a concept recommended in the Google API design guide for some cases (here presented as custom methods with ":customVerb"):
https://cloud.google.com/apis/design/custom_methods

Thank you for your time!

PR #552, opened 2019-11-06
fix: router now support colon sign

Nephylhim:
Useful for custom methods using custom verbs

**Describe the PR**
Add colon sign support for router parse method.

**Relation issue**
https://github.com/swaggo/swag/issues/551

**Additional context**
Custom methods as described in the Google API design documentation (https://cloud.google.com/apis/design/custom_methods) does not work ATM. ex:
/objects:move [post]

Unit test are passing. Fix works well on my side.
Do you want me to change the commit type from "fix" to "feat"?
Is the unit test enough?

Have a good day!

### DIFF
--- a/operation.go
+++ b/operation.go
@@ -437,7 +437,7 @@ func parseMimeTypeList(mimeTypeList string, typeList *[]string, format string) e
 	return nil
 }
 
-var routerPattern = regexp.MustCompile(`([\w\.\/\-{}\+]+)[^\[]+\[([^\]]+)`)
+var routerPattern = regexp.MustCompile(`^(/[\w\.\/\-{}\+:]+)[[:blank:]]+\[(\w+)]`)
 
 // ParseRouterComment parses comment for gived `router` comment string.
 func (operation *Operation) ParseRouterComment(commentLine string) error {
--- a/operation_test.go
+++ b/operation_test.go
@@ -120,7 +120,30 @@ func TestParseRouterCommentWithPlusSign(t *testing.T) {
 	assert.Equal(t, "POST", operation.HTTPMethod)
 }
 
-func TestParseRouterCommentOccursErr(t *testing.T) {
+func TestParseRouterCommentWithColonSign(t *testing.T) {
+	comment := `/@Router /customer/get-wishlist/{wishlist_id}:move [post]`
+	operation := NewOperation()
+	err := operation.ParseComment(comment, nil)
+	assert.NoError(t, err)
+	assert.Equal(t, "/customer/get-wishlist/{wishlist_id}:move", operation.Path)
+	assert.Equal(t, "POST", operation.HTTPMethod)
+}
+
+func TestParseRouterCommentNoColonSignAtPathStartErr(t *testing.T) {
+	comment := `/@Router :customer/get-wishlist/{wishlist_id}:move [post]`
+	operation := NewOperation()
+	err := operation.ParseComment(comment, nil)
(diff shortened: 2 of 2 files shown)


operation.go
```diff
@@ -437,7 +437,7 @@ func parseMimeTypeList(mimeTypeList string, typeList *[]string, format string) e
 	return nil
 }
 
-var routerPattern = regexp.MustCompile(`([\w\.\/\-{}\+]+)[^\[]+\[([^\]]+)`)
+var routerPattern = regexp.MustCompile(`([\w\.\/\-{}\+:]+)[^\[]+\[([^\]]+)`)
```
ubogdan: Please fix the regexp accordigly 
The following routes should not be valid 
"/@Router :customer/get-wishlist/{wishlist_id}:move [post]"
[path format](https://tools.ietf.org/html/std66#section-3.3) 
```
In addition, a URI reference (Section 4.1) may be a relative-path reference, in which case the first path segment cannot contain a colon (":") character.
```

Nephylhim:
Original regex: `([\w\.\/\-{}\+]+)[^\[]+\[([^\]]+)`

Well, I've come up with 4 different solutions, all of them pass the tests (explanation are just after):
1. `^(/[\w\.\/\-{}\+:]+)[^\[]+\[([^\]]+)`
1. `^(/[\w\.\/\-{}\+:]+)[[:blank:]]+\[(\w+)]`
1. `^([\w\.\/\-{}\+]+:?[\w\.\/\-{}\+:]*)[^\[]+\[([^\]]+)`
1. `^([\w\.\/\-{}\+]+(:[\w\.\/\-{}\+:]+)?)[^\[]+\[([^\]]+)`

Now the explanations:
1. It's the solution that **change the less characters** to the original regex
    * `^` at the start of the regex to ensure the first character is correct
    * `(/[\w\.\/\-{}\+:]+)` instead of `([\w\.\/\-{}\+]+)` will ensure that the `/` will always be the first character. After that, `:` is added to the range of accepted characters
    * The rest of the regex haven't been changed
2. It's the solution that **change the most** the regex. But I do think it helps it be **clearer** and **respects more** the documentation.
    * `^(/[\w\.\/\-{}\+:]+)`: same explanations as just above
    * `[[:blank:]]+` instead of `[^\[]+`: use whitespaces or tabs as separation instead of any non `[` characters. Your documentation specify to use spaces as separators (https://github.com/swaggo/swag#api-operation).
    * `\[(\w+)]` instead of `\[([^\]]+`: capture the word between the `[]` characters instead of taking every characters possible but a `]`. Actually, this comment works: `/@Router /api/{id}|,*[get`. A HTTP method should be a word, so `\w` is supposed to do the job.
3. It's the solution that **modify the less the original behavior**. However, it weighs down the regex a lot. 
    * `^` at the start of the regex to ensure the first character is correct
    * `([\w\.\/\-{}\+]+:?[\w\.\/\-{}\+:]*)` instead of `([\w\.\/\-{}\+]+)`. `:?` allows a `:` to exists after a path and `[\w\.\/\-{}\+:]*` allows to continue the URI
    * The rest of the regex haven't been changed
4. An alternative to the third solution. A little bit more explicit but, as go's regex doesn't support non-capturing groups, it forces to **change (the code)** `httpMethod` to `matches[3]` and ensure that `len(matches) != 4` instead of `3`

I'll update my MR with the **second solution** as I do think it is the best one.
I'll also add a bonus test case.

I'll update the MR if you think another solution is better.
Bests

PR #566, opened 2019-11-21
routerPattern should be follow RFC3986

unak:
**Describe the PR**
Current `routerPattern` does not match the path specification described in RFC3986.
OpenAPI 2.0 does not say any about acceptable characters except `{` and `}` (template characters).
Therefore, `routerPattern` should accept all non-reserved characters and should not accept reserved characters described in RFC3986.

**Relation issue**
none

**Additional context**
nothing

### DIFF
--- a/operation.go
+++ b/operation.go
@@ -437,7 +437,7 @@ func parseMimeTypeList(mimeTypeList string, typeList *[]string, format string) e
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


ubogdan:
@unak your PR comes from no open issue. 
More than that your PR doesn't pash the tests. 
Please open an ISSUE to discuss your proposal/issue.

Issue #483, opened 2019-07-29
Doesn't working with import alias for struct fields


dsxack:
**Describe the bug**
When we add import alias for custom type of struct field with this custom type that is not working.

**To Reproduce**
Put the struct with custom type from other package and add alias for this custom type:
```go
package build

import (
       	domainapp "domain/application"
)

type ListResponse struct {
	BuildInfos []domainapp.BuildInfo `json:"data"`
	Err        error                 `json:"-"`
}
```

And some handler with comment
```go
// @Success 200 {object} build.ListResponse
```

Run yaml:

    swag init -g ./main.go --parseDependency --parseVendor

Result:
```yml
build.ListResponse:
    properties:
      data:
        type: string
    type: object
```

**Expected behavior**
Expected result yaml:
```yml
build.ListResponse:
    properties:
      data:
        items:
          $ref: '#/definitions/application.BuildInfo'
        type: array
    type: object
```

**Your swag version**
swag version v1.6.2

**Your go version**
go version go1.12.5 linux/amd64

dsxack:
I can make PR if this issue will be accepted.

ubogdan:
I will be happy to see this issue finally fixed.  (it was "solved" few time in the past #242,  #294)

alexanderbez:
Any updates on this? I'm having an issue where I have two structs defined in a package A and those two structs both have internal types of different packages with the same name, hence I need to use an alias. But aliases don't get captured by swag :-/

marbar3778:
@dsxack do you still have interest in committing a PR to fix this, this is currently blocking us as well

alexanderbez:
I would be happy to look into opening a PR in fixing this. Is there any reference or guide into getting started here? What should I be looking at to update?

Was https://github.com/swaggo/swag/pull/298 closed solely due to lack of activity and unit testing?

dsxack:
@marbar3778 yes, thanks to remember, I'm going to make PR for it tomorrow.

dsxack:
@marbar3778, @alexanderbez, fix is merged into master. Could you check it in your project?

alexanderbez:
Yes, we will report back!

marbar3778:
Tested it out and ran into a similar issue as before. The aliasing works, but it seems to ignore the aliased name of the type. If two types share the same name, in our case `types.Params` is aliased, then in the swagger file it shows as the aliased name but incorrect type definition.

dsxack:
[...]

PR #573, opened 2019-12-03
fix issue with import aliases (#483)

dsxack:
**Describe the PR**
Add the ability to parse refs with import aliases.

**Relation issue**
https://github.com/swaggo/swag/issues/483


ubogdan:
Great job ! Thanks for your contribution !

Issue #581, opened 2019-12-12
map not perfectly supported in struct


sdghchj:
As mentioned by #284 , I tried map members in struct, but not perfect according to [https://swagger.io/docs/specification/data-models/dictionaries/](https://swagger.io/docs/specification/data-models/dictionaries/)

code :+1: 
```
package api

type Child struct {
	Name string
}

type Parent struct {
	//Test1 map[string]interface{}  //not support map[string]interface{} now
	Test2 map[string]string
	Test3 map[string]*string
	Test4 map[string]Child
	Test5 map[string]*Child
}

// @Success 200 {object} Parent
// @Router /api/{id} [get]
func Test(){
}
```
after being parsed:

```
{
   "api.Child": {
      "type": "object",
      "properties": {
         "name": {
            "type": "string"
         }
      }
   },
   "api.Parent": {
      "type": "object",
      "properties": {
         "test2": {
            "description": "Test1 map[string]interface{}",
            "type": "object",
            "additionalProperties": {
               "$ref": "#/definitions/string"   //wrong, string is primitive type
            }
         },
         "test3": {
            "type": "object",
            "additionalProperties": {
               "$ref": "#/definitions/string" //wrong, string is primitive type
            }
         },
         "test4": {
            "type": "object",
            "additionalProperties": {
               "$ref": "#/definitions/Child" //wrong, no package name
            }
         },
         "test5": {
            "type": "object",
            "additionalProperties": {
               "$ref": "#/definitions/Child" //wrong, no package name
            }
         }
      }
   }
}
```

this right doc should be:
```
{
   "api.Child": {
      "type": "object",
      "properties": {
         "name": {
            "type": "string"
         }
      }
   },
   "api.Parent": {
      "type": "object",
      "properties": {
         "test2": {
            "description": "Test1 map[string]interface{}",
            "type": "object",
            "additionalProperties": {
               "type": "string"
            }
         },
         "test3": {
            "type": "object",
            "additionalProperties": {
               "type": "string"
            }
         },
         "test4": {
            "type": "object",
            "additionalProperties": {
               "$ref": "#/definitions/api.Child"
            }
         },
         "test5": {
            "type": "object",
            "additionalProperties": {
               "$ref": "#/definitions/api.Child"
            }
         }
      }
   }
}
```

PR #583, opened 2019-12-13
fully support map and optimization

sdghchj:
1. optimization: parseTypeExpr return poionter to spec.Schema.
2. optimization: struct parsed as definition ref, which may decrease output size.
3. fully support map type, including map[string]interface{},map[string]primitive, also fix bug described in #581.

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -602,7 +602,7 @@ func (parser *Parser) ParseDefinition(pkgName, typeName string, typeSpec *ast.Ty
 	if err != nil {
 		return err
 	}
-	parser.swagger.Definitions[refTypeName] = schema
+	parser.swagger.Definitions[refTypeName] = *schema
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


parser.go
```diff
@@ -648,28 +648,42 @@ func fullTypeName(pkgName, typeName string) string {
...
 		}
-		return parser.swagger.Definitions[refTypeName], nil
+		return &spec.Schema{
+			SchemaProps: spec.SchemaProps{
+				Ref: spec.Ref{
+					Ref: jsonreference.MustCreateRef("#/definitions/" + refTypeName),
+				},
+			},
+		}, nil
+		/*schema := parser.swagger.Definitions[refTypeName]
```
easonlin404: please remove the comment out.

sdghchj: @easonlin404 have removed

parser_test.go
```diff
@@ -253,8 +254,13 @@ func TestGetAllGoFileInfo(t *testing.T) {
 	err := p.getAllGoFileInfo(searchDir)
 
 	assert.NoError(t, err)
-	assert.NotEmpty(t, p.files["testdata/pet/main.go"])
-	assert.NotEmpty(t, p.files["testdata/pet/web/handler.go"])
+	if runtime.GOOS == "windows" {
```
easonlin404: Don't need to specify os for CI.

sdghchj: @easonlin404 I known that, however,it is necessary for some developers preferring windows IDEs  like me, to run these tests first before comitting.  Are you sure to remove it?

easonlin404: In my point, I think the better way is to use [filepath](https://golang.org/pkg/path/filepath/) to determine instead of using `runtime.GOOS`. 😄

easonlin404: and thanks for your miner fix. @sdghchj

easonlin404:
sorry for not promptly reply. @sdghchj Thank you for your contribution. 💯

Issue #562, opened 2019-11-15
Wrong parse of [][]string


art-frela:
**Describe the bug**
When a struct has field(s) of [][]string type, `swag init` makes Unknown Type

**To Reproduce**
1. make struct
```golang
// Content is universal content type, table and raw
type Content struct {
	Headers []string    `bson:"headers" json:"headers" swaggertype:"array,string"`
	Table   [][]string  `bson:"table" json:"table"`
	Raw     interface{} `bson:"raw" json:"raw"`
}
```
2. add this struct to the // @Success description
```golang
// @Success 200 {object} Content
// @Router /api/v2/docs/ [get]
```
3. save and run `swag init`
4. look your output file /path/to/docs.go

```json
"table": {
                    "type": "array",
                    "items": {
                        "type": "\u0026{%!s(token.Pos=1477) \u003cnil\u003e string}"
                    }
                }
```

**Expected behavior**
It's [swagger array](https://swagger.io/docs/specification/data-models/data-types/#array)
that is why docs.go must be
```json
"table": {
                    "type": "array",
                    "items": {
                        "type": "array",
                        "items":{
                            "type": "string"
                        }
                    }
                }
```

**Screenshots**
with error 
![Swagger_wrong](https://user-images.githubusercontent.com/32013768/68943446-4456b280-07bc-11ea-90e6-a3023c4bad24.png)

after hand fix docs.go

![Swagger_correct](https://user-images.githubusercontent.com/32013768/68943460-5173a180-07bc-11ea-876b-72b434c11f10.png)

**Your swag version**
`swag version v1.4.0`

**Your go version**
`go version go1.13.1 darwin/amd64`

**Desktop (please complete the following information):**
 - OS: macOS Catalina v10.15.1
 - Browser: [chrome]
 - Version: [78.0.3904.97]

**Additional context**
In the dockerized app have same behavior

NamelessHH:
My issue is very similar https://github.com/swaggo/swag/issues/584

easonlin404:
fixed by #588 , please help test against the latest master branch.

art-frela:
> fixed by #588 , please help test against the latest master branch.

Thank You!

Manualy tested - all ok.

```json
"content": {
    "headers": [
      "string"
    ],
    "raw": {},
    "table": [
      [
        "string"
      ]
    ]
  },```

tejaskale:
This unfortunately still fails on float 2-D arrays. For eg:  
```go
[...]

Issue #510, opened 2019-09-05
There is a problem with array element support in the structure


singhwang:
**Describe the bug**
// ListAccounts godoc
// @Summary List accounts
// @Description get accounts
// @Tags cluster
// @Accept  json
// @Produce  json
// @Param Authorization header string true " "
// @Param url query string true " "
// @Param start query string true " "
// @Param end query string true " "
// @Param step query string false " "
// @Success 200 {object} common. ResponseData

Add this comment to the function, and then there will be an unsupported situation.

**To Reproduce**
First case 

type Response struct {
	Status  int    `json:"status"`
	Message string `json:"message"`
}

type ResponseData struct {
	Data [][]Response `json: "data"`
}
The error is [&{%!s(token.Pos=143) <nil> Response}].

Second case

type ResponseData struct {
	Data []interface{} `json: "data"`
}
The error is [&{%!s(token.Pos=143) %!s(*ast.FieldList=&{152 [] 153}) %!s(bool=false)}].

Third case

type ResponseData struct {
	Data [][]interface{} `json: "data"`
}
The error is [&{%!s(token.Pos=143) <nil> %!s(*ast.InterfaceType=&{145 0xc0003a2270 false})}]

**Your swag version**
e.g. v1.6.2

**Your go version**
e.g. go1.12.1

**Desktop (please complete the following information):**
 - OS: all
 - Browser: all
 - Version: all

singhwang:
If you can solve the above problems, swag will be more perfect in universal support.

ubogdan:
Swagger is OpenApi compliant , not "universal" api. 

In order to be able to generate request and response objects swagger need to have concrete implementations.

singhwang:
> Swagger is OpenApi compliant , not "universal" api.
> 
> In order to be able to generate request and response objects swagger need to have concrete implementations.
> 
> If you think you can implement the universal support please do a PR.

Thanks for your help.

sdghchj:
try new master branch again

PR #588, opened 2019-12-23
optimize getArrayPropertyName; support [][]string fixing #562

sdghchj:
optimize getArrayPropertyName; 
support [][]string fixing bug mentioned by #562, also #510


easonlin404:
@sdghchj Thanks for your contributions. 👍

PR #614, opened 2020-01-17
Fix dataTypeFormatlink link.

DexterHD:
Fix a link anchor `#dataTypeFormat` which refers nowhere. 
Now it refers to OpenAPI Spec v2, to `Data Types` table.


ubogdan:
Thanks for contributing.

PR #616, opened 2020-01-21
Handle and skip structs with "func" fields.

### LINKED ISSUE #615: Skip structs with `func` type fields

It's not possible to use a structure with `func` type fields as a response or parameter (even if `func` type appears inside the nested field). 

E.g. Imagine a response like this:
```golang
	type SomeObject struct {
		ID int64 `json:"id"`
		Name string `json:"name"`
		IsApplicable   func(id, ppid int64, uid int64) bool `json:"-"`
		AfterIsApplied func(id int64, uid *nt64) `json:"-"`
	}

        type Response struct {
             ID string `json:"id"`
             Value SomeObject `json:"value"`
        }
```

I would expect that `swag` just skip `IsApplicable, AfterIsApplied` fields and generates a proper doc. In fact, it doesn't generate a definition of `SomeObject` because it has `func` fields, but generates a `$ref` link to not existing definition:
```
Response:
    properties:
      name:
        type: string
      value:
        $ref: '#/definitions/SomeObject'
```

As a result, I get an error from swagger, that by specification is invalid:
> Error resolving $ref pointer ".../swagger.yaml#/definitions/SomeObject" Token "SomeObject" does not exist.

I want to use such kinds of structs with skipped `func` fields. 

Prevent generating `$ref` links if the definition wasn't generated.

**Additional context**

-

DexterHD:
**Describe the PR**
Added code for parsing and skipping structure fields with `func` type.

**Relation issue**

fix #615

**Additional context**

-


easonlin404:
@DexterHD Thanks for your contribution.  👍

Issue #640, opened 2020-03-09
need exclude some directories


LinkinStars:
```
├── src
│   ├── a
│     └── api_a.go
│   ├── b
│     └── api_b.go

```

Sometimes I need exclude some dir when generate.   
EX: `swag init -exclude ./src/b`   
This can be very useful when I only want to  export a part of apis.  May be `api_b.go` is Internal API that can't expose to the third party.

sdghchj:
I have tried it. 
"swag init --dir ./src/a" actually does work.

comphilip:
@LinkinStars Are there models shared by both a and b outside a and b folder?

LinkinStars:
@comphilip a and b have their own models.

LinkinStars:
@sdghchj 
If I have some directories `c` `d` `f`...  I can't use "swag init --dir ./src/a ./src/c ./src/d" when I only need to exclude b directory.  Usually, you know, all API is in the same level of directory.

sdghchj:
> @sdghchj
> If I have some directories `c` `d` `f`... I can't use "swag init --dir ./src/a ./src/c ./src/d" when I only need to exclude b directory. Usually, you know, all API is in the same level of directory.

我已经加了，还要等其他人review。如果急需，你可以去我fork的项目里拉下来编译。

sdghchj:
@LinkinStars 
我已经加了，还要等其他人review。如果急需，你可以去我fork的项目里拉下来编译。

Issue #636, opened 2020-03-04
collectionFormat


affanshahid:
Is there a way I can provide the collection format for query params which are suppose to be arrays? The default is `csv` but gin by default works with `multi`

sdghchj:
By description?

affanshahid:
I'm sorry but you could you give me an example? I tried a few things but wasn't able to get it to work

sdghchj:
I mean you can use description to show your API users the format.
```
@param name query []int true "format: name=1&name=2"
```
Swag just generates docs to be showed by gin-Swagger, and now hasn't support collectionFormat for array in query.
Yes, I remember gin-Swagger's default behavior is csv.
If your debugs directly in swagger UI with gin server, it will fail.
Next PR will consider it.
Of course, welcome your PR.

sdghchj:
Try master branch

affanshahid:
Works perfectly in comment form but not working when used in struct tags:
```go
type GetDefinitionMappersQuery struct {
	Name       *string  `json:"name" form:"name"`
	Category   []string `json:"category" form:"category" collectionFormat:"multi"`
	TenantWide *bool    `json:"tenant_wide" form:"tenant_wide"`
}
```

sdghchj:
@affanshahid Said clear in README.md, collectionFormat is only supported for params in query and formdata

pasquale95:
@sdghchj in my case it doesn't work even for formData:

```
type Params struct {
	Status []uint `form:"status" collectionFormat:"multi"` // The Status ids
}
```

Do you know if I'm missing something?

malmeidam:
Still not working

KianYang-Lee:
I think it should be explicitly stated that it is not supported for struct tag, the way it states this now  is a bit misleading

sdghchj:
> [@sdghchj](https://github.com/sdghchj) in my case it doesn't work even for formData:
> 
> type Params struct {
> 	Status []uint `form:"status" collectionFormat:"multi"` // The Status ids
> }
> Do you know if I'm missing something?

Try the global settings in the main file:
```
@query.collection.format multi
```
This will set the default way.

sdghchj:
> I think it should be explicitly stated that it is not supported for struct tag, the way it states this now is a bit misleading

@KianYang-Lee Try latest commit

PR #642, opened 2020-03-14
Support colletionFormat for array params in query; Exclude directories and files

sdghchj:
**Describe the PR**
feature: support collectionFormat for array params in query.
feature: use  flag -exclude to exclude directories and files when searching APIs.
Fix bug: can't find dependent packages when running on windows and both --dir and --parseDependency are set.

**Relation issue**
#636 #640


easonlin404:
@swaggo/swaggo-core-team please help review it, thanks.

Issue #659, opened 2020-04-07
ParseComment error


poccariswet:
**Describe the bug**
swag init error in example/celler
```
ParseComment error in file controller/examples.go :can not find schema type: "controller.integer"
```

**To Reproduce**
1. Go to example/celler.'
2. 'swag init'
3. See error

**Your swag version**
latest

**Your go version**
e.g. 1.14

sdghchj:
int better

PR #660, opened 2020-04-07
[FIX] parse comment error integer to string

poccariswet:
**Describe the PR**

fixed this error
```
ParseComment error in file controller/examples.go :can not find schema type: "controller.integer"
```

**Relation issue**
#659

**Additional context**
n


sdghchj:
int is better

Issue #567, opened 2019-11-22
panic: runtime error: invalid memory address or nil pointer dereference 


nfyxhan:
**Describe the bug**
panic: runtime error: invalid memory address or nil pointer dereference when I use `swag init --parseDependency`

**To Reproduce**
```
export GO111MODULE=on
swag init --parseDependency
```
```
2019/11/22 12:04:01 Generate swagger docs....
2019/11/22 12:04:01 Generate general API Info, search dir:./
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x843d12]

goroutine 1 [running]:
github.com/swaggo/swag.(*Parser).getAllGoFileInfoFromDeps(0xc0002bce00, 0xc0005bda30, 0x0, 0x0)
	/root/pkg/mod/github.com/swaggo/swag@v1.6.3/parser.go:1323 +0x62
github.com/swaggo/swag.(*Parser).getAllGoFileInfoFromDeps(0xc0002bce00, 0xc00098f980, 0x46, 0x0)
	/root/pkg/mod/github.com/swaggo/swag@v1.6.3/parser.go:1341 +0x13e
github.com/swaggo/swag.(*Parser).getAllGoFileInfoFromDeps(0xc0002bce00, 0xc0007529a0, 0x0, 0x0)
	/root/pkg/mod/github.com/swaggo/swag@v1.6.3/parser.go:1341 +0x13e
github.com/swaggo/swag.(*Parser).getAllGoFileInfoFromDeps(0xc0002bce00, 0xc0007bb238, 0x0, 0x0)
	/root/pkg/mod/github.com/swaggo/swag@v1.6.3/parser.go:1341 +0x13e
github.com/swaggo/swag.(*Parser).getAllGoFileInfoFromDeps(0xc0002bce00, 0xc0005a9390, 0x0, 0x0)
	/root/pkg/mod/github.com/swaggo/swag@v1.6.3/parser.go:1341 +0x13e
github.com/swaggo/swag.(*Parser).getAllGoFileInfoFromDeps(0xc0002bce00, 0xc0007cd658, 0x0, 0x0)
	/root/pkg/mod/github.com/swaggo/swag@v1.6.3/parser.go:1341 +0x13e
github.com/swaggo/swag.(*Parser).getAllGoFileInfoFromDeps(0xc0002bce00, 0xc0009594f8, 0x0, 0x0)
[...]

PR #662, opened 2020-04-08
Skip cgo package

MaitreDede:
**Describe the PR**
Swag tries to parse CGO packages, this PR adds a way to handle this.

**Relation issue**
https://github.com/swaggo/swag/issues/567


ubogdan:
Can you please add some tests for it. 
Thanks.

MaitreDede:
Hi,
I have added the unit test. Without the code in `parser.go` it fails. With the code in `parser.go` it passes.

ubogdan:
As far as I understand from the fix, we are not interested in any struct declarations inside CGO packages?

sdghchj:
So big a sample just for cgo?

MaitreDede:
> As far as I understand from the fix, we are not interested in any struct declarations inside CGO packages?

Yes, I think would be too much problems to scan "cgo"... I think the purpose of cgo is only to use non-go libraries, not to expose directly native structs to json...

> So big a sample just for cgo?

Copy/paste of simple ref example, just added some dummy cgo.

sdghchj:
I mean can you compress it after you copy/paste it?

MaitreDede:
Hi,

> I mean can you compress it after you copy/paste it?

What do you ben by "compress" ? Remove redundant files/code ? Or something more ? :)

sdghchj:
> Hi,
> 
> > I mean can you compress it after you copy/paste it?
> 
> What do you ben by "compress" ? Remove redundant files/code ? Or something more ? :)

Remove redundant files/code.

sdghchj:
> Hi,
> 
> > I mean can you compress it after you copy/paste it?
> 
> What do you ben by "compress" ? Remove redundant files/code ? Or something more ? :)

Just one gin router is enough. I think.

MaitreDede:
I think I have it :)

PR #668, opened 2020-04-10
Fix issue with generating docs for embedded structures with alias instead of package name

ShoshinNikita:
**Issue**

For example, there's a package `github.com/user/repo/common` which contains type `Common`:

```go
package common

type Common struct {
    ID int `json:"id"
}
```

If we embed this type in another struct but use an alias instead of the real package name, **swag** will generate wrong documentation:

```go
package main

import (
    c "github.com/user/repo/common"
)

type Response struct {
    c.Common
    Name string `json:"name"
} 
```
Generated documentation:

```json
{
    "definitions": {
        "main.Response": {
            "type": "object",
            "properties": {
                "name": {
                    "type": "string"
                }
            }
        }
    }
}
```

**Solution**

We can suppose that `pkgName` is a package alias, if [`typeSpec`](https://github.com/swaggo/swag/compare/master...ShoshinNikita:master#diff-f615844d3497ff38db57e459d6ef657bR861) is `nil`. So, we can just iterate through import aliases and check can we get non-nil `typeSpec`

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -880,12 +880,8 @@ func (parser *Parser) parseStructField(pkgName string, field *ast.Field) (map[st
 
 			switch schemaType {
 			case "object":
-				if len(schema.SchemaProps.Properties) > 0 {
-					for k, v := range schema.SchemaProps.Properties {
-						properties[k] = v
-					}
-				} else if schema.AdditionalProperties != nil {
-					properties[typeName] = *schema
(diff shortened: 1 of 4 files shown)


sdghchj:
It is also need to be considered that the embeded field here might be a alias type, for example:
```
type Common []string
```
Since a lot of work to do, I refactored a branch `refactored`. 
@easonlin404 What's your idea?

Issue #666, opened 2020-04-10
non-exported struct fields are being exported 


tcarreira:
**Describe the bug**
If I have not-exported fields on a struct, it gets exported as a json example. It should stay hidden

**To Reproduce**

If you have this code:
```
// TestStruct Test Struct
type TestStruct struct {
	tableName    struct{}
	anotherField string
	ID           uint64 `json:"-"`
	Name         string `json:"name"`
	Age          int    `json:"age"`
}

// TestFunction Test
// ---
// @Summary Test Function
// @Description Test non-exported fields
// @Produce json
// @Success 200 {object} TestStruct "Test Struct"
// @Router /test [get]
func TestFunction() {}
```
We get the struct TestStruct with example :
```
{
  "age": 0,
  "anotherField": "string",
  "name": "string",
  "tableName": {}
}
```

**Expected behavior**
we should get:
```
{
  "age": 0,
  "name": "string"
}
```

**Screenshots**
![Screenshot_20200410_051256](https://user-images.githubusercontent.com/5046551/78975628-dc966080-7aea-11ea-96f3-00f4e42dafd1.png)

**Your swag version**
```
$ swag -v
swag version v1.6.5
```

**Your go version**
```
$ go version
go version go1.14.1 linux/amd64
```

**Additional context**
The fields disappear it I explicitly disable them with a json tag:
```
// TestStruct Test Struct
type TestStruct struct {
	tableName    struct{} `json:"-"`
	anotherField string   `json:"-"`
	ID           uint64   `json:"-"`
	Name         string   `json:"name"`
	Age          int      `json:"age"`
}
```

**PS**
Point me in the right direction, and I may get some time to submit a PR

PR #675, opened 2020-04-14
fix: parseField skips not-exported (encoding/json-like)

tcarreira:
**Describe the PR**
In order to match encoding/json behaviour for Marshaling struct into JSON, we should not parse non-exported fields (aka: fields starting with a lower letter)

**Relation issue**
https://github.com/swaggo/swag/issues/666

**Additional context**
With this commit, I had to apply a small fix to a test. I really think it's ok, but please, verify it.

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -1165,6 +1165,11 @@ func (parser *Parser) parseField(pkgName string, field *ast.Field) (*structField
 		return &structField{name: ""}, nil
 	}
 
+	// Skip non-exported fields.
+	if !ast.IsExported(field.Names[0].Name) {
+		return &structField{name: ""}, nil
+	}
+
 	structField := &structField{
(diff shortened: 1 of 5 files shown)


sdghchj:
So you dropped backward compatibility, though it's not a big trouble.

sergeydobrodey:
This feature is very useful when structures are generated using new version of protoc-gen-go: https://blog.golang.org/protobuf-apiv2

Generated structures have technical fields without json:"-" annotation. Maybe you can add run flag if backward compatibility is obligatory?

easonlin404:
Although causing non-backward compatibility, it makes sense that not-exported fields don't need to parse for marshaling struct into JSON.

easonlin404:
What's your guys thought on it? @swaggo/swaggo-core-team

Issue #665, opened 2020-04-10
Example value for struct does not respect tag `json:",string"`


tcarreira:
**Describe the bug**
If you add the `json:",string"` tag to a struct field, the Response example value does not respect it.
 

**To Reproduce**

Create a struct that tags an int as `json:",string"` 
```
type MyUser struct {
    Name string `json:"name" example:"John Doe"`
    Age  int    `json:"age,string" example:"25"`
}
``` 

and document some API endpoint with
```
// @Success 200 {object} MyUser "My User"`
``` 

Then open the Swagger UI and check the return example

The example given is:
```
{
  "age": 25,
  "name": "John Doe"
}
```
25 is not quoted but the actual Response is.

**Expected behaviour**

The Response example value should be 
```
{
  "age": "25",
  "name": "John Doe"
}
```

**Screenshots**
![Screenshot_20200409_211757](https://user-images.githubusercontent.com/5046551/78951286-fe222880-7aa7-11ea-867d-fe3f208091a6.png)

**Your swag version**
$ swag --version
swag version v1.6.5

**Your go version**
$ go version
go version go1.14.1 linux/amd64

sdghchj:
I see.
json package uses `json:",string"` to change field of golang built-in type to string.
PR, welcome.

PR #677, opened 2020-04-15
issue-665: fix example when using `json=",string"`

tcarreira:
**Describe the PR**
Fix the json example when a struct field has the `json:",string"` tag.

**Relation issue**
https://github.com/swaggo/swag/issues/665

**Additional context**
encoding/json states:

> The "string" option signals that a field is stored as JSON inside a JSON-encoded string. It applies only to fields of string, floating point, integer, or boolean types. This extra level of encoding is sometimes used when communicating with JavaScript programs:

---

Before this fix, the provided example would fail json.Unmarshal:
Screenshot_20200414_215316

After fix:
Screenshot_20200414_215541

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -1207,8 +1207,14 @@ func (parser *Parser) parseField(pkgName string, field *ast.Field) (*structField
 	}
 
 	jsonTag := structTag.Get("json")
+	hasStringTag := false
 	// json:"tag,hoge"
 	if strings.Contains(jsonTag, ",") {
+		// json:"name,string" or json:",string"
+		if strings.Contains(jsonTag, ",string") {
+			hasStringTag = true
+		}
+
 		// json:",hoge"
 		if strings.HasPrefix(jsonTag, ",") {
 			jsonTag = ""
@@ -1249,11 +1255,16 @@ func (parser *Parser) parseField(pkgName string, field *ast.Field) (*structField
 		}
 	}
 	if exampleTag := structTag.Get("example"); exampleTag != "" {
-		example, err := defineTypeOfExample(structField.schemaType, structField.arrayType, exampleTag)
-		if err != nil {
-			return nil, err
+		if hasStringTag {
+			// then the example must be in string format
+			structField.exampleValue = exampleTag
+		} else {
+			example, err := defineTypeOfExample(structField.schemaType, structField.arrayType, exampleTag)
+			if err != nil {
+				return nil, err
+			}
+			structField.exampleValue = example
 		}
-		structField.exampleValue = example
 	}
 	if formatTag := structTag.Get("format"); formatTag != "" {
 		structField.formatType = formatTag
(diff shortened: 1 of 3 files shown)


parser.go
```diff
@@ -1207,8 +1207,14 @@ func (parser *Parser) parseField(pkgName string, field *ast.Field) (*structField
 	}
 
 	jsonTag := structTag.Get("json")
+	hasStringTag := false
 	// json:"tag,hoge"
 	if strings.Contains(jsonTag, ",") {
+		// json:"name,string" or json:",string"
+		if strings.Contains(jsonTag, ",string") {
+			hasStringTag = true
```
sdghchj, line 1215: I think one line here is enough
```
structField.schemaType = "string"
```

tcarreira: Hi @sdghchj .

Indeed, that's the fix. (line 1364)
But in order to have working JSON examples, the other stuff is needed. 
Let me clarify this change:

1. identify `json=",string"` tag [1213-1217]
1. if the example tag exists, do not convert from the string, just keep it [1258-1261]
1. in the end (so, I will keep processing tags like min, max, etc... which depend on the original type) 
1.1 I fix the bug [1364]
1.1 If no example is manually set before, I force an example (otherwise, we would have "string" as an example for an int, which would not marshal as well)

Let me know if you need me to clarify anything else. 
I tried to keep these clarifications in the code, but I'm the only one inside my head, so help me to make it clearer

sdghchj: In my opinion, ",string" is like tag `swaggertype:"string"`, so one himself has to make sure the original type can be converted to string directly or by implementing some interface, and know clearly how to define examples,default values, etc. , however, swag can't guarantee it. 
As for tags min,max etc. , they base on the new type specified by tag swaggertype, so it may be a little confusing if they bases on the original type in the case of ",string",  I think.

As you known,",string" only affects integer,string,number,boolean,  if the original type is something else, lines [1258-1261] and [1366-1371] will drop examples. 
 
I prefer to use defer to do some work at the end of a function if strong related codes cross too long to the end. That will looks graceful.

I am not sure whether it is suitable to force an example only in the case of tag ",string".

tcarreira: I understand you, but let me give you just one last argument. 
If you disagree, let me know and I will simplify my PR with your input.

My goal was to make sure you can use the swagger-ui with minimum stress. 
So, in my opinion, when you "try it out" with the defaults, you should get a default behaviour as well. It should "just work".

The default behaviour with `json=",string"` is (un)marshaling as it was the original type.
So, when you "try it out", swagger-ui will force an example on you with invalid values unless you force it somehow. 
I don't agree with this being like tag `swaggertype:"string"` because then you are already focusing on swagger (you will probably set a working example anyway), which is not the case with the `,string` tag.

Maybe the `parser.go` is the wrong place for this functionality (point me to the right place then), but if you don't set some example, an `int` will always fail to unmarshal with a "string".

sdghchj: OK then

PR #681, opened 2020-04-20
Sort query parameters generated from object

breeswish:
**Describe the PR**

Currently swag may generate arbitrary order parameters when parameters come from query & object, like:

```golang
type Student struct {
	Name string
	Age int
	Teachers []string
	SkipField map[string]string
}

// @Param request query Student true "query params"
```

This results in different swagger spec being generated each time even if source code is not modified. Additionally, some openapi client generators like typescript-axios generates each query parameter as function parameters, like:

```typescript
testGet(age?: number, name?: string, teachers?: Array<string>, options?: any): (axios?: AxiosInstance, basePath?: string) => AxiosPromise<void> {
    ...
}
```

Thus, unstable parameter order in the specification file will cause code failed to compile at all.

This PR makes parameter order stable by sorting them before output.

**Relation issue**

**Additional context**


sdghchj:
Changes to be done.

foxluqi:
i think use order tag would be better than field name

breeswish:
@foxluqi Indeed! However currently the order of the tag is lost after being inserted to the hashmap.. I don't have ideas how to do it.

foxluqi:
> @foxluqi Indeed! However currently the order of the tag is lost after being inserted to the hashmap.. I don't have ideas how to do it.

the tag information can be read from schema.Properties?

sdghchj:
@foxluqi  @breeswish  refer to [https://github.com/go-openapi/spec/pull/116](https://github.com/go-openapi/spec/pull/116) and [https://github.com/go-openapi/spec/pull/117](https://github.com/go-openapi/spec/pull/117)

Issue #691, opened 2020-05-01
Add default response annotation


jancona:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to define a default response for an operation, as described here: https://swagger.io/docs/specification/2-0/describing-responses/#default

**Describe the solution you'd like**
In addition to `@produce` and `@error` add a similar new tag `@default` that generates the default response definition.

**Describe alternatives you've considered**
I don't see an alternative other than generating all possible responses for every operation.

**Additional context**
If you think this would be valuable, I can put a PR together.

sdghchj:
It's valuable, PR is welcome

easonlin404:
@jancona Hi, https://github.com/swaggo/swag/pull/837 support default response by using _default_ keyword.

PR #692, opened 2020-05-02
Add support for @default annotation to add a default response type for operations

jancona:
**Describe the PR**
Add support for @default annotation to add a default response type for operations

**Relation issue**
#691 

**Additional context**
Let me know if you think it needs additional tests or documentation.


sdghchj:
Since swagger UI still uses http status code to distinguish success and failure,  `@success` and `@failure` are only meaningful to API developers。

The following is still can be parsed correctly by current version swag.
```
//@success 200 {object} model.Sample
//@success 400 {object} model.Sample
```

I think, reserving `@success` and `@failure` for compatibility, adding `@response` is better. Reference bellow.
```
//@response 200 {object} model.Sample
//@response 400 {object} model.Sample
//@response default {object} model.Sample
``` 

What's your idea?

ubogdan:
@jancona can you please update the PR so the tests will not fail anymore?

easonlin404:
https://github.com/swaggo/swag/pull/837 already supported the functionality.

PR #727, opened 2020-06-06
Refactor based on master branch

sdghchj:
**Describe the PR**
1. This PR is due to refactor master branch, mainly to fix the issue:  two or more models from different packages have the same name as below:

api1.go
 ```
import xxx/yyy/response

response.Order
```

api2.go
```
import aaa/bbb/response

response.Order
```
api3.go
```
import response aaa/bbb/ccc

response.Order
```
In this case, master branch will be ambiguous. This PR will rename them.

2. Use an alias to import a package and the package name is different from its directory name:

mymodel.go in package path "github.com/xxx/yyy/model"
```
package mymodel

type Foo struct{
}
```

main.go
```
package main
import model_alias github.com/xxx/yyy/model

model_alias.Foo
```
In this case, master branch will go to find model.Foo, of course , not exist, mymodel.Foo is right.

3. Fix other issues of master branch:
#682 set multiple searching dirs
#691 add default response
#700 parse alias of alias
#718 parse alias of models

**Side effects**
This PR uses a more strict way to search for model by type name. Before searching the top dir, call `go list` to get its package-import-path, then prefixed to its sub dirs as sub-package-import-path. When looking for a model definition by a type name with a package name, it searches the imports of current ast.File for the correct package, so there are side effects:
1. Since calling `go list` ,It may be a little inefficient relative to master branch, but acceptable.
2. `go list` will fail if there is no go files in working dir, so fix of issue #432 will be erased.
3. `go list` may fail in GOPATH mode.
4. As it searches for models from the imports of current ast.File, if you use a model name from another package as your API request or response, you have to import that package path explicitly. Of course, no need for models in current package.

### CODE REVIEW

**Comment by sdghchj** [comment]
@easonlin404 take a look

File: README.md [hunk 1]
```diff
@@ -264,11 +264,9 @@ func (c *Controller) ShowAccount(ctx *gin.Context) {
 // @Accept  json
 // @Produce  json
 // @Param q query string false "name search by q"
-// @Success 200 {array} model.Account
+// @Response 200 {array} model.Account
```
easonlin404, line 267: Why change to `@Response` tag?

gen/gen.go
```diff
@@ -36,7 +36,7 @@ func New() *Gen {
 
 // Config presents Gen configurations.
 type Config struct {
-	// SearchDir the swag would be parse
+	// SearchDir the swag would be parse,comma separated if multiple
```
easonlin404, line 39: If we want to support search muli directories. I think adding a new one flag is better than using the original flag.

easonlin404:
@sdghchj Thanks for taking the time on it. I know this PR tends to resolve ambiguous package naming and for me. But these changes are too diverge with the latest master branch.

Could you rebase with master and pull small requests? Gradually refactor so that we can easily review each change. 👍

README.md
sdghchj: Refer to #692 , I added a @response tag, the same effect as @success and @failure.
There is already an example showing @success and @failure just above this one.

sdghchj:
> @sdghchj Thanks for taking the time on it. I know this PR tends to resolve ambiguous package naming, but these changes are too diverge with the latest master branch.
> 
> Could you rebase with master and pull small requests? Gradually refactor so that we can easily review each change. 👍

Yes, I have been aware of it. If needed, it can be version 2.0.
Splitting it into smaller pull requst seems to be a harder work  , to be gradually achieved.

gen/gen.go
genofire: I need this feature - is it implemented somewhere (in any other developer branch)?

PR #732, opened 2020-06-10
Add operationId uniqueness check

### LINKED ISSUE #731: Add "operationId" uniqueness check
## Is your feature request related to a problem? Please describe.
Due to [Open API v2 specification](https://swagger.io/specification/v2/#operationObject), `operationId` field inside `Operation Object` must be unique:
> The id **MUST** be unique among all operations described in the API.

Right now `swag` does not have this check which can cause generating the wrong specification.

## Describe the solution you'd like

My proposal is to add this uniqueness check, so `swag` will fail if it finds some `operationId` intersection.

## Describe alternatives you've considered

We can make it a "WARNING"  but at the same time, I would say it's not the right solution because specification clearly says **MUST** be unique.

idexter:
## Describe the PR

Add `operationId` uniqueness check, in accordance with [OpenAPI v2 Specification](https://swagger.io/specification/v2/#operationObject)

## Relation issue

Closes #731

## Additional context

I added a uniqueness check **after the parsing phase**, so it shows a path when an error happens (E.g. `GET /foo/bar`) but did not show the file where duplication happens. 

I did that in this way because I did not succeed adding check inside `ParseRouterAPIInfo()`. The reason I did not succeed is that because when you are going through this range:

https://github.com/swaggo/swag/blob/67cb0975876f155725732d1aeb87fb8310e6aaf9/parser.go#L171-L175

`parser.files` can contain the same file twice. Once with absolute path and once with a relative path. So in some situations, errors happens even if there are no duplications in `operationId`


idexter:
> however, operationsIds is better to be in a local scope.

@sdghchj I tried to move it inside the `checkOperationIDUniqueness()` function. Is that thing you was talking about? 

https://github.com/swaggo/swag/pull/732/files#diff-f615844d3497ff38db57e459d6ef657bR1508-R1521

sdghchj:
> > however, operationsIds is better to be in a local scope.
> 
> @sdghchj I tried to move it inside the `checkOperationIDUniqueness()` function. Is that thing you was talking about?
> 
> https://github.com/swaggo/swag/pull/732/files#diff-f615844d3497ff38db57e459d6ef657bR1508-R1521

That's it

easonlin404:
@idexter Thanks for your contributions.

PR #736, opened 2020-06-13
Refactor3:  fix more than one models have the same the from different packages

sdghchj:
1. fix: more than one models have the same the from different packages
2. fix: parse alias of alias of ……

I have to say, I try to changes the least code,  but still a lot of work,  and especially repair those tests.


easonlin404:
@sdghchj Although this change is very a lot, I can understand the purpose you want to fix though the testing case. Thanks for the effort you're trying to keep changes the least. Awesome man. 💯

Issue #738, opened 2020-06-16
err when parse api.md if got err as below when got double quotation mark in api.md


wkj89:
err as below when got double quotation mark in api.md
`Errors
Hide
Parser error on line 5
missed comma between flow collection entries`

api.md file looks like:
`
"a"
`

I guess:

var SwaggerInfo = swaggerInfo{
	Version:     "0.1",
	Host:        "",
	BasePath:    "/",
	Schemes:     []string{},
	Title:       "Cmdb Api",
	Description: "something",
}

I think should chnage     Description: "something"  to   Description: `something`
user `` instead of "" or  escape  ”“

sdghchj:
More detailed, please.

wkj89:
> More detailed, please.

well，I got an api.md file for the main function with description.markdown comment

And in api.md file only have :    "a"

then  swag init

run main.go   and   open http://127.0.0.1:8080/swagger/index.html

i got err:

```
Errors 
Hide Parser error on line 5 
missed comma between flow collection entries

```

sdghchj:
Got it, It's a bug of swagger ui. I think.

j0s:
This problem can also be reproduced without any markdown involved. Just type any string with quotation marks in the general API `@Description` field.
I actually this is a bug stemming from the two step application of templates in  [`writeGoDoc()`](https://github.com/swaggo/swag/blob/940d07268e91254eb18c2978cda233381ab643cb/gen/gen.go#L160) and then in [`ReadDoc()`](https://github.com/swaggo/swag/blob/940d07268e91254eb18c2978cda233381ab643cb/gen/gen.go#L283), resulting in the escaping of quotes disappearing. Since the problem only appears when running the generated doc.go, it doesn't show up in `swagger.{json,yaml}`. I came up with a workaround that seems to work well enough: https://github.com/swaggo/swag/pull/743

wkj89:
> This problem can also be reproduced without any markdown involved. Just type any string with quotation marks in the general API `@Description` field.
[...]

PR #743, opened 2020-07-06
Support quotes and tabs in general API description

j0s:
**Describe the PR**
Handle quotes and tabs in general API description. The two-stage application of templates when generating `docs.go`, and then applying a template in [`ReadDoc()`](https://github.com/swaggo/swag/blob/940d07268e91254eb18c2978cda233381ab643cb/gen/gen.go#L283), resulted in escaping of quotation mark getting lost. This PR re-adds adds those quotations using a less than beautiful workaround, but at least it works on my test cases.
It also adds a test that compiles the generated file `docs.go`, runs the `ReadDoc()` method, and compares the output to a template.

**Related issue**
https://github.com/swaggo/swag/issues/738

**Additional context**
I stumbled upon this when expanding my documentation using markdown. When writing long descriptions, supporting quotation marks and tabs would be very useful.


easonlin404:
@j0s sorry, because ci check had some issue there before. could you help retry the build again? eg: some minor changes

j0s:
Ok, so I fixed a linting error. But the coverage must be caused by codecov not being run for a while right? It's complaining that coverage of `parser.go` has decreased, but I haven't touched that file?

j0s:
Is there anything blocking this from being merged, except for the travis job that does not seem to run?

j0s:
I would still like this to be merged if it looks good to you.

jecolasurdo:
Any movement on this?

sdghchj (approved):

Issue #761, opened 2020-07-26
Maximum and Minimum attributes do not function


parnic:
**Describe the bug**
Using the attributes `maximum` or `minimum` in swagger documentation does not emit the max/min output.

**To Reproduce**
1. Create api documentation enforcing the `maximum` value of a field.
2. Run `swag init`
3. Observe maximum is not present.

**Expected behavior**
Maximum is exported.

**Your swag version**
1.6.7

**Your go version**
1.14.6

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Brave (Chromium)
 - Version: 1.11.101 Chromium: 84.0.4147.89 (Official Build) (64-bit)

**Additional context**
This applies to both the Maximum and Minimum attributes. Some documentation says these need to be `maxinum` and `mininum` instead.

sdghchj:
So this is an error of spelling ? 
Ha , I thought it is combined with **maxi**mum and **num**ber

parnic:
Well it's not my project, but I've never seen "maxinum" anywhere before, so I assumed it was an error. Swagger itself uses maximum/minimum: https://swagger.io/docs/specification/data-models/data-types/

PR #762, opened 2020-07-26
Fix typos in maximum/minimum attributes

parnic:
Fixes "maximum" needing to be "maxinum" to work, and "minimum" needing to be "mininum" to work.

Retains support for the old field names and adds tests to ensure the old names continue to work.

**Relation issue**
Fixes https://github.com/swaggo/swag/issues/761

### DIFF
--- a/operation.go
+++ b/operation.go
@@ -274,11 +274,13 @@ func (operation *Operation) ParseParamComment(commentLine string, astFile *ast.F
 var regexAttributes = map[string]*regexp.Regexp{
 	// for Enums(A, B)
 	"enums": regexp.MustCompile(`(?i)\s+enums\(.*\)`),
-	// for Minimum(0)
+	// for maximum(0)
 	"maxinum": regexp.MustCompile(`(?i)\s+maxinum\(.*\)`),
-	// for Maximum(0)
+	"maximum": regexp.MustCompile(`(?i)\s+maximum\(.*\)`),
+	// for minimum(0)
 	"mininum": regexp.MustCompile(`(?i)\s+mininum\(.*\)`),
-	// for Maximum(0)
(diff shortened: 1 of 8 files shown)


operation.go
```diff
@@ -303,13 +303,13 @@ func (operation *Operation) parseAndExtractionParamAttribute(commentLine, object
 			if err != nil {
 				return err
 			}
-		case "maxinum":
+		case "maximum":
```
Nerzal, line 306: I think u can just use 

```go
 case "maxinum":
 case "maximum":
        // rest of code here
```

instead of having the code 2 times :)

parnic: I'm not sure what you mean. Go's case statements don't work like that, they're comma separated, but this is switching on the key of the regex, which I modified to accept either spelling, I didn't do anything two times.

parnic: To be clear, I didn't add anything, I just modified the existing framework. The "maximum" here is a switched value on the key of the `regexAttributes` map above. The only actual required change to fix this is the modification of the regexes on lines 278 and 280, the rest are making the remaining code consistent in terms of spelling. If you would prefer to keep the typos in these case statements, that can be done, but I just figured as long as I was fixing the typo, might as well fix it everywhere.

Nerzal: oh i'm sorry i missread that. guess i shouldn't review PRs before my first coffee! sorry again!

parnic:
> It looks good for me too. But I think we need to update readme as well. And maybe need a change log to highlight it's not compatible with the old swag version.
> 
> @parnic could you help update the readme as well?

The readmes are updated in this pr and it is still compatible with the old version. Both the typo and the correct spelling will work.

Issue #300, opened 2019-02-14
Support objects for formData


lsattem:
**Is your feature request related to a problem? Please describe.**
Not that i know of.

**Describe the solution you'd like**
I have not seen a solution to this in the documentation, but the documentation on @Param is a bit lacking (i.e. does not mention formData being an option).
When setting @Param, it would be nice if you could specify object and object type. This way it's clear what sort of input my endpoint expects. For example, **the following go code...**
```golang
package main

type MyStruct struct{
	AnInt int
}

// @Summary test form data
// @Accept multipart/form-data
// @Param payload formData main.MyStruct true "form data"
// @Router /test [post]
func test(){}
```

**...provides the following swagger:**
```json
{
    "info": {
        "contact": {},
        "license": {}
    },
    "paths": {
        "/test": {
            "post": {
                "consumes": [
                    "multipart/form-data"
                ],
                "summary": "test form data",
                "parameters": [
                    {
                        "type": "main.MyStruct",
                        "description": "form data",
                        "name": "payload",
                        "in": "formData",
                        "required": true
                    }
                ]
            }
        }
    }
}
```
What **I would like** is for the swagger to include the type def, and look like **this**:
```json 
{
    "info": {
        "contact": {},
        "license": {}
    },
    "paths": {
        "/test": {
            "post": {
                "consumes": [
                    "multipart/form-data"
                ],
                "summary": "test form data",
                "parameters": [
                    {
                        "type": "main.MyStruct",
                        "description": "form data",
                        "name": "payload",
                        "in": "formData",
                        "required": true
                    }
                ]
            }
        }
    },
    "definitions": {
        "main.MyStruct": {
            "type": "object",
            "properties": {
                "anInt": {
                    "type": "integer"
                }
            }
        }
    }
}
```
**Describe alternatives you've considered**
An alternative, but ugly, solution is to add a dummy function that contains the struct as an object, such as this one.
```golang
// @Failure 200 {object} main.MyStruct "wish this wasn't necessary"
func dummy(){}
```

I could also use param type 'body', but that is also somewhat ugly (and misleading).

**Additional context**
N/A

pei0804:
I see.
I'm thoughting for solution.

JayenLee:
I have the same problem


type equipmentQuery struct {
	ProductName string `json:"product_name" form:"product_name"` // 产品名称
	Sn          string `json:"sn" form:"sn"`                     // sn号
[...]

PR #770, opened 2020-08-08
feat: generate formData params via go struct

BLumia:
**Describe the PR**
Allow generate params for formData by using go struct.

**Relation issue**
Intended to resolve https://github.com/swaggo/swag/issues/300

**Additional context**
`go test` passed even if I didn't change anything related to the unit test, it seems the old test doesn't cover this case. I didn't add any extra test cases for this PR since I'm not sure about how to write unit test for this case. Hope that's okay.

btw if this PR can get merged, the documentation may also need to update.

### DIFF
--- a/operation.go
+++ b/operation.go
@@ -167,12 +167,12 @@ func (operation *Operation) ParseParamComment(commentLine string, astFile *ast.F
 	param := createParameter(paramType, description, name, refType, required)
 
 	switch paramType {
-	case "path", "header", "formData":
+	case "path", "header":
 		switch objectType {
 		case ARRAY, OBJECT:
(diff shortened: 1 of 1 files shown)


birdycn:
How to use

BLumia:
> How to use

The same as when using with `query`

birdycn:
> > How to use
> 
> The same as when using with `query`

```
type Params struct{
  File *multipart.FileHeader `form:"file"
}

How to write tag

BLumia:
Please consider [Read The Fantastic Manual](https://github.com/swaggo/swag#example-value-of-struct).

btw, if you still got issues, I suggest you open a new issue thread instead of reply in the merged PR.

Issue #793, opened 2020-09-19
Error in data model fails to parse


lahsivjar:
**Describe the bug**
If error type is a field in a data model for request or response then it fails to parse with an error "cannot find type definition: error"

**To Reproduce**
1. Create a struct with error field
2. Use the above struct to define a a swagger annotation 
3. Run swag init

**Expected behavior**
No error for field with error type

**Your swag version**
1.6.7 or master

**Your go version**
e.g. 1.13.14

sdghchj:
error is an interface.
If you take it as string, `swaggertype:"string"` may help, I think.

lahsivjar:
This should work, but my problem is that error is a part of a struct from a third-party package. Any idea how I can add the swaggertype to such struct?

ubogdan:
The only way is to import the package in a vendor folder and apply a patch to it ...

PR #794, opened 2020-09-19
Add support for error field in model

lahsivjar:
**Describe the PR**
Add support for error type ina a field in a data model

**Relation issue**
https://github.com/swaggo/swag/issues/793

**Additional context**
 If error type is a field in a data model for request or response then it fails to parse with an error "cannot find type definition: error"


ubogdan (approved):
LGTM

ubogdan:
The output of json.Marshal for the `error` type is an empty object `{}`. 

```go
package main

import (
	"encoding/json"
	"errors"
	"fmt"
)

type Data struct {
	Name  string
	Age   int
	Error error
}

func main() {
	data := Data{
		Name:  "John",
		Error: errors.New("unable to parse user John"),
	}
	json, err := json.Marshal(data)
	if err != nil {
		panic(err)
	}
	fmt.Printf("json output: %s", json)
}
```

Output:
```
json output: {"Name":"John","Age":0,"Error":{}}
```

The `error` type is not a Primitive type is a private Interface type and the field should be skipped during struct parsing. 
 
@lahsivjar Please let me know if you are willing to update the PR according to the conclusion above.

Thanks,
Bogdan

ubogdan (changes requested):
Error is a private interface, it shouldn't be marshaled to string. See comment above.

lahsivjar:
@ubogdan A bit busy this week, will take a look and try to submit a PR next week.

ubogdan:
@lahsivjar No worries. Take your time!

ubogdan:
@lahsivjar any updates here?

ubogdan:
@lahsivjar I will like to include this in the current release, unfortunately, it's not ready to be released.

lahsivjar:
@ubogdan Have been away from the project which uses this for a long time. Looking at the above PR isn't what you expect just a revert of [this commit](https://github.com/swaggo/swag/pull/794/commits/de3d6b4f5acbecbec4c0fca8b38231794c348c02)?

ubogdan:
Ok then, you can close the PR.

PR #806, opened 2020-10-13
fix: honor x-order while sorting query parameters

lantw44:
**Describe the PR**
Instead of always sorting parameters with name, which can surprise users
because x-order is ignored, sort them with the slice type declared in
github.com/go-openapi/spec to reuse its comparison method.

**Relation issue**
https://github.com/swaggo/swag/pull/681

**Additional context**
It can be simplified if https://github.com/go-openapi/spec/pull/117 is merged.


sdghchj (approved):
LGTM

PR #835, opened 2020-11-18
Update dependencies and fix tests

akojo:
**Describe the PR**

Upgrade github.com/go-openapi/spec and golang.org/x/tools to latest
versions.

Upgrading github.com/go-openapi/spec v0.19.11 -> v.0.19.14 generates
empty "description" members for responses and empty "authorizationUrl"
members for securityDefinitions when they're missing. This causes tests
to fail. Fix by adding the missing empty definitions to expected test
results.

**Additional context**
Since travis-ci test builds always fetch deps and build before running tests,
they break because of the change in go-openapi/spec

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,14 +5,16 @@ require (
 	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751
 	github.com/ghodss/yaml v1.0.0
 	github.com/gin-gonic/gin v1.6.3
-	github.com/go-openapi/spec v0.19.11
+	github.com/go-openapi/spec v0.19.14
 	github.com/go-openapi/swag v0.19.11 // indirect
 	github.com/go-playground/validator/v10 v10.4.1 // indirect
 	github.com/gofrs/uuid v3.3.0+incompatible
 	github.com/golang/protobuf v1.4.3 // indirect
 	github.com/json-iterator/go v1.1.10 // indirect
 	github.com/mailru/easyjson v0.7.6 // indirect
+	github.com/shopspring/decimal v1.2.0
 	github.com/stretchr/testify v1.6.1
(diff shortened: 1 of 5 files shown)


sdghchj:
I just can't figure out why it generates empty  members?

akojo:
According to the [spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md) `description` and `authorizationUrl` are required fields in those contexts. Apparently that's the reason they are now always included

sdghchj:
> According to the [spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md) `description` and `authorizationUrl` are required fields in those contexts. Apparently that's the reason they are now always included

So where is it that the member `description` is set required?

akojo:
> So where is it that the member `description` is set required, I mean in source code?

[Here's](https://github.com/go-openapi/spec/commit/119312bce69c7a518c3382167495c53240bb7966) in spec the commit that always generates those two members. Apparently it's included in version 0.19.13 and onwards

sdghchj:
> > So where is it that the member `description` is set required, I mean in source code?
> 
> [Here's](https://github.com/go-openapi/spec/commit/119312bce69c7a518c3382167495c53240bb7966) in spec the commit that always generates those two members. Apparently it's included in version 0.19.13 and onwards

OK I see.
But, as you memtioned that it's included in version 0.19.13,  however, I just upgrade it to v0.19.12, it still failed when running those tests.  Refer to [https://github.com/swaggo/swag/pull/836/checks?check_run_id=1422025199](https://github.com/swaggo/swag/pull/836/checks?check_run_id=1422025199)

sdghchj (approved):
LGTM

sdghchj:
@akojo Thank you all the same.

Issue #843, opened 2020-11-26
Swaggertype object second parameter


mlesar:
**Describe the bug**
After [this commit](https://github.com/swaggo/swag/commit/c9056f0363b524c7d1e36f675a3784bf80151e98) you can't have `swaggertype` defined as `swaggertype:"object"` which is an issue if you want to accept anything on that property.

**To Reproduce**
```go
type Request struct {
	Type int             `json:"type"`
	Data json.RawMessage `json:"data" swaggertype:"object"`
}
```

Error:
```
need object item type after object
```

**Expected behavior**
Before the mentioned commit you would get this inside the definition:
```json
...
"data": {
    "type": "object"
}
...
```

**Your swag version**
v1.6.9

**Your go version**
1.14

**Desktop (please complete the following information):**
 - OS: macOS

Is there a way to bypass this to not require a second property when using `object` for `swaggertype`?

sdghchj:
type Request struct {
	Type int             `json:"type"`
	Data json.RawMessage `json:"data" swaggertype:"string"`
}

mlesar:
> type Request struct {
> Type int `json:"type"`
> Data json.RawMessage `json:"data" swaggertype:"string"`
> }

Why would I use `string` if an actual object is sent?

sdghchj:
OK， I got it.
Could you please commit a PR to fix it?

Modify [here](https://github.com/swaggo/swag/blob/master/schema.go#L174)

as

```
case OBJECT:
		if len(types) == 1 {
			return PrimitiveSchema(types[0]), nil
		}
		schema, err := BuildCustomSchema(types[1:])
		if err != nil {
			return nil, err
		}
		return spec.MapProperty(schema), nil
```

mlesar:
>Could you please commit a PR to fix it?

Sure, here it is: https://github.com/swaggo/swag/pull/844

PR #844, opened 2020-11-27
Optional additional parameter when using object in swaggertype

mlesar:
**Describe the PR**
Not requiring an additional parameter to be set when using `object` in `swaggertype`.

**Relation issue**
https://github.com/swaggo/swag/issues/843


sdghchj (approved):

sdghchj:
Thank you

PR #861, opened 2020-12-28
@category tag allowing to split definitions

huttarichard:
**Motivation**

- At larger scale when you have more than just one application more than just one API server, you might face collisions. Eg. `/v1/clients` mounted in binary `cmd/store` might collide with `/v1/clients` mounted in `cmd/admin` (for example). 
- Sometimes It make sense to just filter some endpoints out of definition due to fact endpoints might have sensitive nature. In this case for example the `cmd/admin` might contain some sensitive stuff you do not want to expose. Would be better to have two schemes where one is for admin one for store.
- For sure having multiple services being mounted at root of url without prefix is common. That means scopes of these services differ that means multiple schemes are required.

**Proposed Solution**

In this PR I added `@category` tag which allows to filter definitions. Doesn't affect performance and nothing is practically changed (for existing users).

It will allow to compile whole project as a one piece which helps to prevent naming collision for data structures which are commonly used across all packages. 

Eg you can do this:

```
// In endpoint
// @category store [...list of comma separed categories]
```

```
//go:generate swag init --category store,common --output cmd/store/docs  -g swagger.go
//go:generate swag init --category admin,common --output cmd/admin/docs  -g swagger.go
```

Note, tags in this matter are not very helpful as some users (guessing here, that is what we do) use them to group for example resources together, which would mean you would in command line had to specify all resources, that sounds to me annoying.

### DIFF
--- a/README.md
+++ b/README.md
@@ -370,6 +370,7 @@ When a short string in your documentation is insufficient, or you need images, c
 | description.markdown     |  A short description of the application. The description will be read from a file named like endpointname.md| // @description.file endpoint.description.markdown  |
 | id          | A unique string used to identify the operation. Must be unique among all API operations.                                   |
 | tags        | A list of tags to each API operation that separated by commas.                                                             |
+| category    | A list of categories for given endpoint. Useful if you want to split your definitions.                                     |
 | summary     | A short summary of what the operation does.                                                                                |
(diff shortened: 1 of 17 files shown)


ubogdan:
@huttarichard sorry for bothering you, but is this related to an open issue ?

Issue #872, opened 2021-01-20
execute command `goreportcard-cli -v` will output errors 


windvalley:
**Describe the bug**
When  I use `goreportcard-cli -v` command to check the code quality of my Go project, the following error will be output:
```text
Grade: A+ (99.6%)
Files: 73
Issues: 1
gofmt: 100%
go_vet: 98%
	router/init.go
		Line 21: error: package project-foo/docs is not in GOROOT (/usr/local/Cellar/go/1.15/libexec/src/project-foo/docs) (vet)
gocyclo: 100%
golint: 100%
ineffassign: 100%
license: 100%
misspell: 100%
```

**To Reproduce**
1. install goreportcard-cli 
```bash
brew tap alecthomas/homebrew-tap
brew install gometalinter
go get github.com/gojp/goreportcard/cmd/goreportcard-cli
```
2. `swag init` to generate dir `docs/`
3.  `goreportcard-cli -v` at Go project root dir
4. See the errors above

**Expected behavior**
```text
Grade: A+ (100.0%)
Files: 74
Issues: 0
gofmt: 100%
go_vet: 100%
gocyclo: 100%
golint: 100%
ineffassign: 100%
license: 100%
misspell: 100%
```

**Your swag version**
v1.7.0

**Your go version**
go version go1.15 darwin/amd64

ubogdan:
Merged.

PR #873, opened 2021-01-20
Fix the issue that using 'goreportcard-cli -v' will output errors (swaggo#872)

windvalley:
**Describe the PR**
Fix the issue that using 'goreportcard-cli -v' will output errors (swaggo#872)

**Relation issue**
https://github.com/swaggo/swag/issues/872


gen/gen.go
```diff
@@ -249,10 +249,9 @@ func (g *Gen) writeGoDoc(packageName string, output io.Writer, swagger *spec.Swa
 	return err
 }
 
-var packageTemplate = `// GENERATED BY THE COMMAND ABOVE; DO NOT EDIT
+var packageTemplate = `// Package docs GENERATED BY THE COMMAND ABOVE; DO NOT EDIT
 // This file was generated by swaggo/swag{{ if .GeneratedTime }} at
```
ubogdan, line 253: since the package name is configurable via ` {{.PackageName}}`, please update it accordingly.

ubogdan (changes requested):

gen/gen.go
windvalley: > since the package name is configurable via ` {{.PackageName}}`, please update it accordingly.

OK, thanks

ubogdan (approved):

ubogdan:
@windvalley Thanks for your contribution.

PR #885, opened 2021-02-15
Allow different names for swagger specs

joshk0:
In the case where we need to include multiple docs packages in the same
process, swag.Register will fail. Despite looking like having support
for swagger specs with different names it actually does not work.

Define RegisterName and ReadDocName functions which allow you to specify
the name of the swag. Default method will still use this path but with
the 'classic' swag name of 'swagger'.

**Describe the PR**
e.g. add cool parser.

**Relation issue**

**Additional context**


joshk0:
The CI fails for the same reason as the [latest master commit](https://travis-ci.org/github/swaggo/swag/jobs/759229920).

ubogdan:
We need to investigate and rollback asap. Please be patient.

joshk0:
Hi, is this waiting for me to add tests to restore the code coverage?

aleiphoenix:
+1 to this. I want to serve multiple swagger instances from one gin router.  Currently swag.Register could only called once, no matter what name parameter is.

This will require some changes in gin-swagger `CustomHandler` too.

ubogdan:
> Hi, is this waiting for me to add tests to restore the code coverage?

Yes. You introduced new code branches that are not covered by testing.

swagger.go
```diff
@@ -22,20 +23,33 @@ type Swagger interface {
...
 }
 
 // ReadDoc reads swagger document.
 func ReadDoc() (string, error) {
-	if swag != nil {
+	return ReadDocName(Name)
+}
+
+// ReadDocName reads swagger document with specific name.
+func ReadDocName(name string) (string, error) {
```
ubogdan, line 47: Changing the function signature is not an acceptable solution. This is a breaking change and breaks all other projects.
`func ReadDoc() (string, error) {`
to 
`func ReadDocName(name string) (string, error) {`

swagger.go
```diff
@@ -22,20 +23,33 @@ type Swagger interface {
...
+	if swags == nil {
+		return "", errors.New("no swag has yet been registered")
+	} else if swag, ok := swags[name]; !ok {
```
ubogdan, line 50: Race condition: searching in the `swags` map without a mutex !

swagger.go
```diff
@@ -22,20 +23,33 @@ type Swagger interface {
...
+		return "", fmt.Errorf("no swag named \"%s\" was registered", name)
+	} else {
```
ubogdan, line 52: please use return instead if nested else.

gen/gen.go
```diff
@@ -309,6 +309,6 @@ func (s *s) ReadDoc() string {
 }
 
 func init() {
-	swag.Register(swag.Name, &s{})
+	swag.Register({{ printf "%q + %q" .Host .BasePath}}, &s{})
```
ubogdan, line 312: What is the expected behavior In the event of many API applications don't have a Host and Basepath defined?

ubogdan (changes requested):
We can't introduce breaking changes.

ubogdan:
Closing: fixed in #1022 !

PR #892, opened 2021-03-01
lint code

ubogdan:
**Describe the PR**
[x] Lint code
[x] Refactor annotations, struct tags to constant

**Relation issue**
NONE

**Additional context**
I will like to do more refactor but I'm afraid the review will never be done or it will require too much time.


ubogdan:
@easonlin404 looks like Travis is getting stuck for some reason ...

Issue #905, opened 2021-03-25
Now that we have exclude, maybe we also need to include?


nimohunter:
same to #640 , if the path tree like this:
```
├── src
│   ├── package_a
│     └── someapi_of_a
│   ├── package_b
│     └── someapi_of_b
│   ├── package_c
│     └── someapi_of_c1
│     └── someapi_of_c2
```

i just want need generate swagger file in  package_c/someapi_of_c2, we have to use 

--exclude src/package_a,src/package_b/;src/package_c/ someapi_of_c1 

Can it be optimized here? 

**Describe the solution you'd like**
if this is useful and meaning, i'll try to pull request.

nimohunter:
i try to use `swag init -d ./src/package_c/someapi_of_c2` ,but this "dir value" is not suitable for this situation absolutely.

nimohunter:
or maybe we need a new arg like searchroot or something ?

nimohunter:
analysis the code, find the reason why  return error like "cannot find type definition: xxxxxx" is that search struct define root path is same with Directory you want to parse.  

but if your api request or response define struct not located in the the Directory you want to parse. then will return this error.

nimohunter:
i find struct type define also be ignore by exclude args, so move the skip excloud path function and add include. 
and, if set include, exclude will be invalid

Welcome to discuss this idea :)

sdghchj:
> i find struct type define also be ignore by exclude args, so move the skip excloud path function and add include.
> and, if set include, exclude will be invalid
> 
> Welcome to discuss this idea :)

So what if one wants to exclude some sub directories in the directory included by --includes.

nimohunter:
> > i find struct type define also be ignore by exclude args, so move the skip excloud path function and add include.
> > and, if set include, exclude will be invalid
> > Welcome to discuss this idea :)
> 
> So what if one wants to exclude some sub directories in the directory included by --includes.

good question, i'll check whether my pr can support your idea.

this the reason why i want to add this arg:
in some case, i just want to generate a little api interface(for example, src/package_a/), but some struct is define out of src/package_a/,  the struct is located in src/struct_define. so i add this arg.


├── src
│   ├── package_a
│     └── someapi_of_a
│   ├── package_b
│     └── someapi_of_b
│   ├── struct_define
│     └── some_struct_define

nimohunter:
> > i find struct type define also be ignore by exclude args, so move the skip excloud path function and add include.
[...]

PR #906, opened 2021-03-25
fix: add include, and exclude is invalid if include is set

nimohunter:
**Describe the PR**
add include , if include is set ,exclude path is invalid.

**Relation issue**
#905 

**Additional context**
i find struct type define also be ignore by exclude args, so move the skip excloud path function and add include.  Welcome to discuss this idea


Lehp:
parser_test.go 3385 pretty nice example for your missing tests

Issue #920, opened 2021-04-21
Improve some error messages


vyskocilm:
**Is your feature request related to a problem? Please describe.**
`swag` sometimes produces hard to follow messages, especially if the struct definitions is non trivial

```
2021/04/21 13:04:00 ParseComment error in file src.go :object is unsupported type in example value
```

**Describe the solution you'd like**
The example value shall be printed too, because it helps to identify the problematic definition faster.

```
2021/04/21 13:04:00 ParseComment error in file src.go :object is unsupported type in example value This is a neat example
```

**Describe alternatives you've considered**
Or to print a line number, but that'd be harder to implement

```
2021/04/21 13:04:00 ParseComment error in file src.go:42 :object is unsupported type in example value
```

**Additional context**

PR #921, opened 2021-04-21
provide example and enum values when reporting unsupported type

vyskocilm:
**Describe the PR**
This helps debugging of the problems and improves the usability of swag. As other error messages prints the
problematic value, this seems like the easiest change.

**Relation issue**
https://github.com/swaggo/swag/issues/920


ubogdan (approved):
LGTM

PR #927, opened 2021-04-29
support boolean false value in extension

HurSungYun:
**Describe the PR**
This change supports to use boolean false value as extension values. Currently, extensions field only supports boolean true and string values. However, boolean false value is needed for some frequently used extensions such as `x-omitempty`.

All swagger extensions should start with `x-`, so prefix `!` does not break anything and it can be a good rule.

**Relation issue**

None

**Additional context**

My first approach was like `x-omitempty=false`, but it may break some extensions that uses `false` as string value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -675,7 +675,7 @@ type Account struct {
 
 ```go
 type Account struct {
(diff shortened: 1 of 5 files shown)


sdghchj (approved):
LGTM

Issue #661, opened 2020-04-07
Export filtered APIs with specified tags or other ways


oreo0725:
**Is your feature request related to a problem? Please describe.**
Sometime we have to only export partial APIs to generate the document for external partners.
For example,
I have 2 APIs `/user/new` and `/internal/user/new` in `users_resource.go`

- Only the `/user/new` I wanna export it to a `swagger-public.yaml`
- Both `/user/new` and `internal/user/new` should be exported to another `swagger.yaml`

**Describe the solution you'd like**
Is that possible to have a **--exclude-tags** or **--include-tag** options for the generator function?
or just a whole new annotation to fullfil this requirement 

**Describe alternatives you've considered**
Use `@Tags` may not be a good choice because it will create a new layout group for HTML generated by such **swagger-codegen**

ubogdan:
@oreo0725 you may need to figure out a workaround since swagger doesn't not support this and I don't see it implemented. Since you have 2 APIs , you may use to create 2 endopints http://api.domain.com and http://internal-api.domain com or you can have different roots line /api/new and /internal/new .

oreo0725:
@ubogdan,
Perhaps my explanation is not that clear.
What I really want is like the `--exclude` argument implementation, not only for the file/dir level but for the **API/function level**

ubogdan:
@oreo0725 This is an Open Source project. Please feel free and contribute via PR and we will be more than welcome to add this feature.

igorkova:
@oreo0725 Do you plan on creating PR from your forked branch?

johan-lejdung:
I'm in the same boat. We have a monorepo with both internal and public API's, using the exclude flag doesn't really work at all times (depending on how you've created your package structure). 

It seems https://github.com/avyfinans/swag/pull/1 added the ability to parse by extension (awesome addition!):

Add comment tag like
// @x-public true

on cli use swag init --parseExtension public

result will be that only those operations with @x-public true will be saved to the swagger definition

---

The new flag is still not referenced in the Readme file though. I'd suggest adding that and then closing this PR.

ubogdan:
As the contributor mentioned 

> [igorkova](https://github.com/igorkova) commented [on Jun 2](https://github.com/swaggo/swag/pull/1219#issuecomment-1144746967)
> This is not 100% ready yet. Let us get back

igorkova:
Ok. We are back on this one.

I see that in the meantime we got something similar in #1379, however, that doesn't solve the use-case we have.

[...]

Issue #693, opened 2020-05-03
Swagger Doc - Ignore Controller


grtn316:
**Is your feature request related to a problem? Please describe.**
I have a controller that I would like to hide from the documentation. I have been unable to find a solution in the documentation. It would also be useful to hide individual routes as well.

**Describe the solution you'd like**
Controller - Add a comment at the top of the file to ignore.
Route - // @Tags Hide

**Describe alternatives you've considered**
Only way it appears I can hide it is to remove my existing controller.

sdghchj:
I suggest you put  your individual routes away from your searching dir.

oreo0725:
@grtn316 
I have the same requirement #661

ubogdan:
Let me guess: Security by obscurity ! 

Creating hidden routes doesn't make the software more secure. On the other hand, you shouldn't add comments on the controller you don't want to make visible.

grtn316:
> Let me guess: Security by obscurity !
> 
> Creating hidden routes doesn't make the software more secure. On the other hand, you shouldn't add comments on the controller you don't want to make visible.

Actually, I'd just like APIs only used by the system to be aware of them by hiding them in my public swagger docs. Even if they are "secured", I don't desire bots "testing" my secure endpoints they found in my public swagger docs. It takes resources from my server and is unnecessary when hiding them in the documentation will quickly cut down a lot of that traffic.

I should not have to build another app to host system only APIs when I can hide them with decorations in other programing languages.

PR #952, opened 2021-07-05
if the tag @hidden is in the route annotations it will skip it

ch4rl1e5:
**Describe the PR**
I had a large number of routes and needed some of them just hidden from the swagger.yaml/json file. So I added a new condition for the tag @hidden. If this tag is present, swag will skip that route.
Ex:

```
/ ShowAccount godoc
// @Hidden
// @Summary Show a account
// @Description get string by ID
// @ID get-string-by-int
// @Accept  json
// @Produce  json
// @Param id path int true "Account ID"
// @Success 200 {object} model.Account
// @Header 200 {string} Token "qwerty"
// @Failure 400,404 {object} httputil.HTTPError
// @Failure 500 {object} httputil.HTTPError
// @Failure default {object} httputil.DefaultError
// @Router /accounts/{id} [get]
func (c *Controller) ShowAccount(ctx *gin.Context) {
```

In the example above, swag will skip the /accounts/{id} [get]  documentation specifcally

**Relation issue**
https://github.com/swaggo/swag/issues/661

**Additional context**
I found necessary to create a error file and the error ErrIsHidden to return when a route is skipped and identify the error
The tag should be on the top of the annotations for best performance

### DIFF
--- a/cmd/swag/main.go
+++ b/cmd/swag/main.go
@@ -2,12 +2,11 @@ package main
 
 import (
 	"fmt"
-	"log"
-	"os"
-
 	"github.com/swaggo/swag"
 	"github.com/swaggo/swag/gen"
 	"github.com/urfave/cli/v2"
+	"log"
+	"os"
 )
 
 const (
--- a/errors.go
+++ b/errors.go
@@ -1,0 +1,5 @@
+package swag
+
+import "errors"
+
+var ErrIsHidden = errors.New("endpoint is hidden")
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ require (
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-openapi/spec v0.20.3
 	github.com/stretchr/testify v1.7.0
+	github.com/urfave/cli v1.22.5
 	github.com/urfave/cli/v2 v2.3.0
 	golang.org/x/tools v0.1.0
 )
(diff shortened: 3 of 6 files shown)


ubogdan:
@ch4rl1e5 Is this related to any open issue? I find it more simple than simply skipping the swagger comments on top of the functions you don't want to include in the generated file.

ch4rl1e5:
@ubogdan I linked this https://github.com/swaggo/swag/issues/661, bu actually it's has more to do with this https://github.com/swaggo/swag/issues/693

The purpose is just being able to filter what functions will be in the generated file.

ubogdan:
The same behavior can be obtained if you add a new line between the comment and the function declaration. 
```
// ShowAccount godoc
// @Summary Show a account
// @Description get string by ID
// @ID get-string-by-int
// @Accept  json
// @Produce  json
// @Param id path int true "Account ID"
// @Success 200 {object} model.Account
// @Header 200 {string} Token "qwerty"
// @Failure 400,404 {object} httputil.HTTPError
// @Failure 500 {object} httputil.HTTPError
// @Failure default {object} httputil.DefaultError
// @Router /accounts/{id} [get]

func (c *Controller) ShowAccount(ctx *gin.Context) {
```

Your workaround is quite close to the behavior wanted in the referenced issue but I'm still not confident will any of them.

Issue #460, opened 2019-07-18
multiple methods


pjebs:
There seems to be no way to specify multiple HTTP methods for a given route. ie. a given route supports GET & POST.

ubogdan:
Yes, it's true, but Can you do that in swagger editor ?

pjebs:
yes its possible

ubogdan:
ok then, why don't you do a PR for this.

pjebs:
can you show approx area of code where to make the change?

ubogdan:
You can start from here:
https://github.com/swaggo/swag/blob/bd1235d8e7291c9914afe94717e0c3d2972f45bc/operation.go#L106-L109

Please note you will need to write code testing if you want your PR to merged into master.

ubogdan:
If you decide to go with a pr , please mention this issue id.
Thanks.

idefixcert:
> yes its possible

How is that possible?

PR #957, opened 2021-07-24
Add support to specify multiple routes for the same function

pc-coder:
**Describe the PR**
Add support to specify multiple routes for the same function

**Relation issue**
https://github.com/swaggo/swag/issues/460

**Additional context**
In the case of multiple @Router comments, swagger creates routes using information only from the last comment.
This change would help us add swagger routes for each comment.

### DIFF
--- a/operation.go
+++ b/operation.go
@@ -18,11 +18,16 @@ import (
 	"golang.org/x/tools/go/loader"
 )
 
+// RouteProperties describes HTTP properties of a single router comment.
+type RouteProperties struct {
+	HTTPMethod string
+	Path       string
+}
+
 // Operation describes a single API operation on a path.
 // For more information: https://github.com/swaggo/swag#api-operation
 type Operation struct {
-	HTTPMethod string
-	Path       string
+	RouterProperties []RouteProperties
 	spec.Operation
 
 	parser              *Parser
@@ -54,8 +59,8 @@ func NewOperation(parser *Parser, options ...func(*Operation)) *Operation {
 	}
 
 	result := &Operation{
-		parser:     parser,
-		HTTPMethod: "get",
+		parser:           parser,
+		RouterProperties: []RouteProperties{},
 		Operation: spec.Operation{
 			OperationProps: spec.OperationProps{},
 			VendorExtensible: spec.VendorExtensible{
@@ -516,8 +521,11 @@ func (operation *Operation) ParseRouterComment(commentLine string) error {
 	path := matches[1]
 	httpMethod := matches[2]
 
-	operation.Path = path
-	operation.HTTPMethod = strings.ToUpper(httpMethod)
+	signature := RouteProperties{
+		Path:       path,
+		HTTPMethod: strings.ToUpper(httpMethod),
+	}
+	operation.RouterProperties = append(operation.RouterProperties, signature)
 
 	return nil
 }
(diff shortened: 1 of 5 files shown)


operation.go
```diff
@@ -516,9 +522,16 @@ func (operation *Operation) ParseRouterComment(commentLine string) error {
...
-	operation.Path = path
-	operation.HTTPMethod = strings.ToUpper(httpMethod)
+	signature := RouteProperties{
+		Path:       path,
+		HTTPMethod: strings.ToUpper(httpMethod),
+	}
+	if operation.RouterProperties[0].IsDefault {
+		operation.RouterProperties = []RouteProperties{signature}
+		return nil
+	}
```
ubogdan, line 533: So you added the isDefault property instead of using len(operation.RouterProperties)?

operation.go
```diff
@@ -18,11 +18,17 @@ import (
 	"golang.org/x/tools/go/loader"
 )
 
+// RouteProperties describes HTTP properties of a single router comment.
+type RouteProperties struct {
+	HTTPMethod string
+	Path       string
+	IsDefault  bool
```
ubogdan, line 25: I know a boolean doesn't eat too much memory, but you are not using it outside `ParseRouterComment` so please get rid of it.

ubogdan (changes requested):
Except for the 2 comments left LTGTM.

operation.go
pc-coder: I also thought about that initially. But, If I add
` if len(operation.RouterProperties) == 1 && (compare the values of HTTPMethod, Path with default) {
		//override the default
	}`
We won't be able to differentiate in cases when a user explicitly adds the default values in comments and would end up overriding the user entered values.

I can think of another way, if we remove the default initialization
from https://github.com/pc-coder/swag/blob/master/operation.go#L64
and check for len == 0 at https://github.com/pc-coder/swag/blob/master/parser.go#L570 and then add defaults, we can get rid of IsDefault.

Thoughts?

ubogdan: you can initialize operation with an empty slice and simply append the slice unconditionaly
```
result := &Operation{
                ..... snip
		RouterProperties: []RouteProperties{},
}
```

```
// ParseRouterComment parses comment for gived `router` comment string.
func (operation *Operation) ParseRouterComment(commentLine string) error {
	var matches []string

	if matches = routerPattern.FindStringSubmatch(commentLine); len(matches) != 3 {
		return fmt.Errorf("can not parse router comment \"%s\"", commentLine)
	}
	path := matches[1]
	httpMethod := matches[2]

	signature := RouteProperties{
		Path:       path,
		HTTPMethod: strings.ToUpper(httpMethod),
	}

	operation.RouterProperties = append(operation.RouterProperties, signature)

	return nil
}
```

pc-coder: Have made the requested changes.

I thought removing default initialization would be a breaking change.

ubogdan (approved):

ubogdan:
I thin after we document this feature in [README.md](https://github.com/swaggo/swag#api-operation) we are ready to merge.

pc-coder:
> Before we merge this into master we need to document the feature in [README.md](https://github.com/swaggo/swag#api-operation) so all our users will be aware of it.

Have updated the readme with an example.

ubogdan (approved):
LGTM

ubogdan:
Thanks for your contribution!

PR #968, opened 2021-08-06
Handle duplicate route declaration better

rubensayshi:
Fixes issue when you accidentally have `@Router /some/route` declared twice.  

Previously this would cause either of the 2 definitions to be used and it leaves the user very confused.

Tbh I'd personally just prefer the hard error when this mistake is made, but I think for backwards compatibility it might be nice to either have it warn by default (as it is now in this PR) or have the option to turn it into a warning, so I added the `Strict` option flag.  
I think this could also be useful for more checks in the future?  

Additionally I made `RangeFiles` sort the files to make sure it happens in deterministic order, prior to my fix to warn / error when there's a duplicate route and without this you'd randomly get one of the 2 routes and that *really* confuses the hell of you because instead of thinking you made a mistake it just looks like a very weird bug.  

The "fix issue with files being parsed twice" was something that came up when I added the warning about duplicate routes and I noticed that in `TestGen_cgoImports` it was triggering as well, this was because it was parsing `simple_cgo/api` from *both* the searchDirs and from the `ParseDependency` block.

If you'd like me to split the PR into multiple then please let me know, or if you want it without the `Strict` config option for example

### DIFF
--- a/gen/gen.go
+++ b/gen/gen.go
@@ -60,6 +60,9 @@ type Config struct {
 	// ParseInternal whether swag should parse internal packages
 	ParseInternal bool
 
+	// Strict whether swag should error or warn when it detects cases which are most likely user errors
+	Strict bool
+
 	// MarkdownFilesDir used to find markdownfiles, which can be used for tag descriptions
 	MarkdownFilesDir string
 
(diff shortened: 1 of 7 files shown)


sdghchj:
Thank you for your efforts.
If strict is set, it is necessary to sort the files, but whether it is still necessary to sort the files when strict is not set, in consideration of higher performance.

rubensayshi:
if you want I can change that, but I'd argue that the sorting is so lightweight that it's not worth it

sdghchj:
> if you want I can change that, but I'd argue that the sorting is so lightweight that it's not worth it

That depends the scale of a project I think.
Don't bother however.
Leave someone in future to optimize it if needed.

Issue #970, opened 2021-08-09
Support @x- attributes for security definition


pmorelli92:
Seeing how was #452 done, can we get the same functionality for security definitions? 

I would need to generate this:

```
securityDefinitions:
  google_id_token:
    authorizationUrl: ""
    flow: "implicit"
    type: "oauth2"
    x-google-audiences: "1234"
    x-google-issuer: "https://accounts.google.com"
    x-google-jwks_uri: "https://www.googleapis.com/oauth2/v3/certs"
```

But if I write the following code:

```
// @securitydefinitions.oauth2.implicit google_id_token
// @authorizationurl ""
// @x-google-audiences "1234"
// @x-google-issuer "https://accounts.google.com"
// @x-google-jwks_uri "https://www.googleapis.com/oauth2/v3/certs"
```

I get the following snippet where the `@x-attributes` are outside the security definition.

```
securityDefinitions:
  google_id_token:
    authorizationUrl: '""'
    flow: implicit
    type: oauth2
swagger: "2.0"
x-google-audiences: "1234"
x-google-issuer: https://accounts.google.com
x-google-jwks_uri: https://www.googleapis.com/oauth2/v3/certs
```

This should be valid according the `OpenAPI Extensions` document.

PR #971, opened 2021-08-09
Added support for @x- attributes on security definition

pmorelli92:
Relates to issue #970.

Before this PR the only `@x-` attribute supported on the security definition was the hardcoded `x-tokenName`. After this PR all the `@x-` attributes that are parsed in the same block than the security definition are going to be added.

Added extra logic for not adding twice the `@x-` attributes that are present on the security definition and removed the hardcoded case scenario for token name as well.

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -377,15 +377,28 @@ func (parser *Parser) ParseGeneralAPIInfo(mainAPIFile string) error {
 					return err
 				}
 				securityMap[value] = securitySchemeOAuth2AccessToken(attrMap["@authorizationurl"], attrMap["@tokenurl"], scopes, extensions)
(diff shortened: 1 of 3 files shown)


pmorelli92:
@sdghchj @ubogdan  Hi guys! Sorry to bother you with the tag, but I would like to know if you can give some love to this PR. Unfortunately, at the company I am working, I have a deadline soon and I would need to know if I can use this dependency or not. Sorry again

pmorelli92:
Done! @ubogdan

ubogdan (approved):
LGTM

ubogdan:
@pmorelli92 Thanks for your contribution.

PR #972, opened 2021-08-11
chore: release v1.8.0

rubensayshi:
Last release was in December, I think it would be good to do a new release.

Particularly the gin security release would be good to include since it fixes a CVE and some dependency management tools will flag this as an issue.

I was creating an issue to ask for a release, but figured by doing a PR if you agree then all you need to do is click the merge button, sorry if this comes over too aggressive just trying to make your life easier ;)

### DIFF
--- a/version.go
+++ b/version.go
@@ -1,4 +1,4 @@
 package swag
 
 // Version of swag
-const Version = "v1.7.0"
+const Version = "v1.8.0"


rubensayshi:
how can this make the coverage go down o.O

ubogdan:
@rubensayshi Unfortunately we can't jump to 1.8.0 just because of a security fix.

ubogdan:
We just made a release for 1.7.1.

rubensayshi:
:+1: thanks for the release

PR #999, opened 2021-09-23
Add multipleOf field to parser

jimmycannlfs:
**Describe the PR**
Add ability to specify `multipleOf` struct field

**Relation issue**
N/A


ubogdan:
To Review the PR, we need to have a passed test and positive code coverage. 
Unfortunately, we had some trouble with the Travis-ci pipeline, and the build didn't run.

Please do another push with a slight change, so the build is triggered.

jimmycannlfs:
Thanks @ubogdan have pushed a small change to trigger, if you could please unblock the workflow :D

PR #1009, opened 2021-09-28
add for support paramType 'form'

liut:
In the last swagger version, I had a large number of param types that used 'form', and I found that in this version, `form` became `formData`. I think simple is good.

### DIFF
--- a/operation.go
+++ b/operation.go
@@ -221,7 +221,7 @@ func (operation *Operation) ParseParamComment(commentLine string, astFile *ast.F
 		refType = strings.TrimPrefix(refType, "[]")
 		refType = TransToValidSchemeType(refType)
 	} else if IsPrimitiveType(refType) ||
-		paramType == "formData" && refType == "file" {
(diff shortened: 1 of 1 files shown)


ubogdan:
In the following [doc](https://swagger.io/docs/specification/2-0/describing-parameters/#form-parameters) you can find the reason:
```
Also, form parameters cannot coexist with the in:body parameter, because formData is a specific way of describing the body. 
```

ubogdan:
@liut is there any open issue for this?

liut:
I checked 2_0 docs and there is no conflict with using form as an alias for formData. likes:
```yml
    post:
      consumes:
      - application/json
      - application/x-www-form-urlencoded
      description: login
      operationId: api-1-login-post
      parameters:
      - description: Username
        in: form
        name: username
        required: true
        type: string
      - description: Password
        in: form
        name: password
        required: true
        type: string
      produces:
      - application/json
```

Issue #1016, opened 2021-10-04
Crash when generating doc for recursive structure


arunpoudel:
**Describe the bug**
swaggo crashes when generating doc for recursive structure

**To Reproduce**
I have prepared a sample that causes it to crash. Can be found at https://github.com/medsensio/swag-go-crash-demo

**Expected behavior**
swagger documentation should be updated

**Screenshots**
N/A

**Exception**
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x7dc91f]

goroutine 1 [running]:
github.com/swaggo/swag.(*Parser).getRefTypeSchema(0xc0001e6000, 0xc00011c7a0, 0x0)
        /home/fr3ak/go/pkg/mod/github.com/swaggo/swag@v1.7.3/parser.go:798 +0x7f
github.com/swaggo/swag.(*Parser).getTypeSchema(0xc0001e6000, {0xc0001201f0, 0x9}, 0xc8b3a0, 0x1)
        /home/fr3ak/go/pkg/mod/github.com/swaggo/swag@v1.7.3/parser.go:748 +0x96d
github.com/swaggo/swag.(*Parser).parseTypeExpr(0xc0001e6000, 0x98f600, {0x998c28, 0xc00011c180}, 0xb0)
        /home/fr3ak/go/pkg/mod/github.com/swaggo/swag@v1.7.3/parser.go:907 +0xab
github.com/swaggo/swag.(*Parser).parseTypeExpr(0xc0001e6000, 0xc00011a0f0, {0x998778, 0xc00011a0f0}, 0x1)
        /home/fr3ak/go/pkg/mod/github.com/swaggo/swag@v1.7.3/parser.go:920 +0x3f1
github.com/swaggo/swag.(*Parser).parseStructField(0xc0001bec58, 0xc00011abd0, 0xc00012a180)
        /home/fr3ak/go/pkg/mod/github.com/swaggo/swag@v1.7.3/parser.go:1045 +0x405
github.com/swaggo/swag.(*Parser).parseStruct(0xc00011f2b0, 0xc00002abc0, 0xc00011a120)
        /home/fr3ak/go/pkg/mod/github.com/swaggo/swag@v1.7.3/parser.go:952 +0xf0
github.com/swaggo/swag.(*Parser).parseTypeExpr(0xc0001e6000, 0x8f7133, {0x998f58, 0xc0001180a8}, 0xf0)
[...]

PR #1017, opened 2021-10-06
added fix for recursive-structure

arunpoudel:
Added fix for #1016 

***Description***
This specific case is resolved by utilizing `allOf` in swagger 2.0. When a struct is embedded into another struct (Composition), the Compose struct is then parsed and referenced in Composing struct using `allOf`.

This covers my current use-case.


ubogdan:
It looks like this implementation changes the way go embedding was implemented.
```
type Resource struct {
	Element
}
```
Should be rendered as:
```
        "main.Resource": {
           "type": "object",
            "properties": {
                "extension": {
                    "type": "array",
                    "items": {
                        "$ref": "#/definitions/main.Extension"
                    }
                },
                "id": {
                    "type": "string"
                }
            }
        }

arunpoudel:
Should I update other tests to reflect those changes to schema?

arunpoudel:
I don't seem to see proper way of implementing this without breaking old contracts.

Issue #765, opened 2020-08-03
No way to select multiple accepts at UI


sudersen:
**No way to select multiple accepts at UI*
While the documentation comment over the implementation does have a way to supply multiple accepts, it doesn't have a to select all the accepts at once. 

**To Reproduce**
Supply the under for multiple accepts in the code
```
// @Accept json, text/csv
// @Produce json, text/csv
```

**Expected behavior**
At the UI end, I'd be expecting to select multiple accepts as allowed by W3C convention. The UI should be a multi select of accept rather than basing it on the *Content-Type* of the response 

**Screenshots**
<img width="1432" alt="image" src="https://user-images.githubusercontent.com/12055231/89197241-b6ba2d80-d5ab-11ea-9af7-3fc44be5f7b6.png">

As it could be seen, we don't have a way to select multiple accepts which is otherwise allowed by the standards.

ubogdan:
I'm trying to understand the issue but as far as I see there is only 1 accept and 1 produce in your definition.
```
// @Accept json, text/csv
// @Produce json, text/csv
```
Can you be more specific?

ubogdan:
Are u sure ? 
![multiple_consume](https://user-images.githubusercontent.com/4992696/136951131-601f5311-169f-4015-88eb-8487089af501.png)

ubogdan:
The fact the annotations need to be written in a comment line makes this very error-prone.

Issue #1020, opened 2021-10-09
Body parameter should support plaintext


jrschumacher:
**Describe the bug**
When using body parameter and the content type is `text/plain` the parameter `schema.type` should be `string` per [spec](https://github.com/swaggo/swag/blob/9fb19d0a36ed06a9d9b82e65322c7e1dbf143004/operation.go#L315-L320)

**To Reproduce**

```
// Generate Bisac and Keywords
// @Summary Plain text process
// @Description Plain text process
// @Param text body string true "Text to process"
// @Accept text/plain
// @Router /process [post]
```

**Expected behavior**

```
paths:
  /process:
    post:
      consumes:
      - plain/text
      description: Plain text process
      parameters:
      - description: Text to process
        in: body
        name: text
        required: true
        schema:
          type: string
```

**Your swag version**
v1.7.3

**Your go version**
v1.17.1

jrschumacher:
Should probably add a check to ensure we are considering the content type when using `body`
https://github.com/swaggo/swag/blob/9fb19d0a36ed06a9d9b82e65322c7e1dbf143004/operation.go#L315-L320

jrschumacher:
The `parseAPIObjectSchema` doesn't have the right context to make this decision:
https://github.com/swaggo/swag/blob/9fb19d0a36ed06a9d9b82e65322c7e1dbf143004/operation.go#L780-L799

sdghchj:
PR is welcome.

PR #1025, opened 2021-10-12
clarify accept param ussage

ubogdan:
**Describe the PR**
Clarify the usage of @accept param.

**Relation issue**
#765

**Additional context**
None.


sdghchj:
It's amazing you konw Chinese.
Could you please also take a look at #1020, which is a similar issue.

Issue #985, opened 2021-08-25
Support `null` Package and Data-Types


naftulikay:
**Describe the bug**

`swag init` fails when compiling models which contain structs from the `gopkg.in/guregu/null.v4` package.

**To Reproduce**

I have a `PATCH` endpoint where every field is optional, so my struct looks like this:

```golang
import "gopkg.in/guregu/null.v4"

type Request struct {
    Email null.String
    IsAdmin null.Bool
}
```

Having marked up my endpoint like so:

```
// UserUpdateHandler godoc
// ...
// @Param user body models.Request true "Updated user data."
// ...
```

When I run `swag init`, it chokes on items in the `gopkg.in/guregu/null.v4` package:

```
2021/08/25 15:50:45 ParseComment error in file pkg/routes/sadmin/users/root.go :cannot find type definition: null.Bool
```

**Expected behavior**

I expected for `swag` to understand that `null.*` are simply nullable types.

**Screenshots**

**Your swag version**

1.7.1

**Your go version**

```
go version go1.16.6 linux/amd64
```

**Desktop (please complete the following information):**

 - OS: elementary OS Loki (Ubuntu 16.04 LTS) amd64

**Additional context**

If I shouldn't be using the `null.*` times for optional values in models, what _should_ I be using?

SmartPhoneJava:
Try instead of
```go
import "gopkg.in/guregu/null.v4"

type Request struct {
    Email null.String
    IsAdmin null.Bool
}
```

this one 
```go
import null "gopkg.in/guregu/null.v4"

type Request struct {
    Email null.String
    IsAdmin null.Bool
}
```

ubogdan:
@naftulikay, have you figured it out?

naftulikay:
@SmartPhoneJava why would changing the imported package name work? Is there some logic within swaggo that will understand it if I force the package name? The code above does compile.

@ubogdan I haven't tested this yet because I don't see how it would work differently. I'll attempt it at some point. Can you link to code or documentation that covers how this works?

ubogdan:
@naftulikay If the suggested workaround works it means there is an issue in the swag parser related to `strings.Split`. 
Swagger is Open Source. Feel free to look at the source code.

Please test and let us know.

SmartPhoneJava:
@naftulikay swaggo has different strategies to parse named and none named imports for detecting imported types. We encountered the opposite problem.  Named imports did not work when unnamed imports worked without error

naftulikay:
[...]

PR #1059, opened 2021-11-26
feat: support global type overrides

ian-fox:
**Describe the PR**
When using generated files (such as from [sqlboiler](https://github.com/volatiletech/sqlboiler)) we encountered a situation where we wanted to override a type (e.g. have `null.String`, a struct with custom json marshal/unmarshalers, show up in the swagger docs as `string` instead of as a `struct` with a string and a bool) but could not use the `swaggertype` struct tag overrides, as sqlboiler does not allow you to customize those on a per-type basis.

This PR adds a global "overrides file" where you can specify a mapping of what type is actually used in the code, and what type (basic or otherwise) you'd like it to render as in the generated docs.

Please let me know if you think there are any changes that should be made!

**Relation issue**
https://github.com/swaggo/swag/issues/985 seems related

**Additional context**
The solution to our problem seemed to be either modifying sqlboiler to allow it to insert custom tags on a per-type basis (e.g. add a `swaggertype:"string"` tag whenever it generated a struct with a `null.String`) or to modify swaggo to be able to do replacements without the struct tag; this seemed like it was probably more broadly useful because it may help with other generators than just sqlboiler.


ubogdan:
@ian-fox, is there any PR regarding this issue? I'm asking this because we appreciate everyone's effort to write a PR, but sometimes it is safer to validate the solution before writing the implementation.

ian-fox:
There is not, I figured it was small enough that is write it up, learn something in the process, and if it wasn't accepted then that's fine because it didn't take too long anyway.

If there are any changes you'd suggest I'm happy to make them, and if it's just a poor design that you'd rather not accept that's also totally fine.

ubogdan:
@ian-fox the ability to replace the definitions without touching the source code sound good to me. 

Somehow an attempt to solve this issue was partially solved by introducing "swaggerype" a long time ago. 

I'm thinking it would be nice to have a simple dot file in the same way git has".gitingore". YAML/TOML is too heavy (the libraries are not light at all and it will introduce additional unwanted dependencies) and JSON is not so user-friendly when it comes to editing.

Eventually, we can add at a later time the ability to control the flow like cmd params do, without the need to update the CI pipeline.

Example swag dot file (.swag)
``` go
// replace a defintion 
replace database/sql.NullInt64 int64

// skip parsing a package
skip github.com/grpc/grpc-go

```

ian-fox:
Sounds good to me! I'll go add those changes 😄

ian-fox:
I think all the issues with the tests and static analysis should be fixed now.

README.md
```diff
@@ -683,6 +684,33 @@ generated swagger doc as follows:
...
+type MyStruct struct {
+  ID     sql.NullInt64 `json:"id"`
+}
+```
+
+`overrides.yaml`:
+```yaml
+"database/sql.NullInt64": "int"
+```
+
```
ubogdan, line 704: This needs to be updated accordingly.

gen/gen.go
```diff
@@ -179,6 +205,34 @@ func (g *Gen) formatSource(src []byte) []byte {
...
+	overridesReplace := regexp.MustCompile(`replace\s+(\S+)\s+(\S+)`)
+	overridesSkip := regexp.MustCompile(`skip\s+(\S+)`)
+	overridesComment := regexp.MustCompile(`^//.*`)
+	nonWhitespace := regexp.MustCompile(`\S`)
+
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		if parts := overridesReplace.FindStringSubmatch(line); parts != nil {
+			overrides[parts[1]] = parts[2] // parts[0] is the full line
```
ubogdan, line 221: strings.Split should be enough for the beginning (regexp are too expensive)
```
args := strings.Split() 
switch strings.Tolower(arg[0]) {
case "replace":
  // Replace logic goes here
case "skip":
  // Skip logic goes here
case "//":
  continue // skip to next line
}
```

gen/gen.go
```diff
@@ -179,6 +205,34 @@ func (g *Gen) formatSource(src []byte) []byte {
...
+		} else if parts := overridesSkip.FindStringSubmatch(line); parts != nil {
+			overrides[parts[1]] = "" // parts[0] is the full line
+		} else if nonWhitespace.MatchString(line) && !overridesComment.MatchString(line) {
+			return nil, fmt.Errorf("could not parse override: '%s'", line)
+		}
```
ubogdan, line 226: else if statements make code harder to follow. try to use the "fail fast" approach.

gen/gen.go
```diff
@@ -93,11 +101,29 @@ func (g *Gen) Build(config *Config) error {
 		}
 	}
 
+	var overrides map[string]string
+	if config.OverridesFile != "" {
```
ubogdan, line 105: what if use `os.Stat` in order to detect there is the override file (.swagoveride) in the working directory and we try to load and parse instead of waiting for a command line param?

ubogdan (changes requested):

ubogdan:
I'm thinking it's easier to finish the code changes first and after that, you can take care of the code coverage.

gen/gen.go
ian-fox: It already does that. Right now it looks for .swaggo by default and only does something else if the flag is provided, but can easily change that to be .swagoverride.

ubogdan: `.swaggo` is fine

ubogdan:
Can you please fix the failing tests?

ian-fox:
Yep. Sorry about that, was counting on CI to run the monkey-patch related ones because monkey does not work on my computer.

ubogdan:
We need to keep compatible with older GO versions. 
```
gen/gen_test.go:7:2: package io/fs is not in GOROOT (/opt/hostedtoolcache/go/1.15.15/x64/src/io/fs)
```

try simply return errors.New("file does not exist");

ubogdan (approved):
LGTM

ian-fox:
Thank you for all your help!

ubogdan:
@ian-fox Thanks for your contribution. 

You're wellcome.

PR #1073, opened 2021-12-10
chore:remove unnecessary check

nmmsantos:
**Describe the PR**
This will remove the check for existence of the `SearchDir` on `swag fmt` which is breaking the comma separated list feature from `swag fmt -d cmd/hello,./`, fore example.

**Relation issue**

**Additional context**
The check is being made after splitting the list on the comma character, later, in the `FormatAPI` call.

### DIFF
--- a/format/format.go
+++ b/format/format.go
@@ -1,9 +1,7 @@
 package format
 
 import (
-	"fmt"
 	"log"
-	"os"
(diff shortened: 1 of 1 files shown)


ubogdan (approved):
LGTM

PR #1094, opened 2021-12-31
Refactor some things for the needs of github.com/go-chai/chai

e-nikolov:
**Describe the PR**

I am working on https://github.com/go-chai/chai - an extension for the chi router that provides support for type safe http handlers via generics. This allows it to also provide swagger spec generation that automatically detects the request/response types, http methods and route paths. Chai uses swaggo/swag annotations for the parts of the swagger spec that cannot be automatically inferred.

I am using swaggo/swag as a library but some of the functions that I needed were unexported and others I needed to modify a bit. I forked swag and made the changes I needed there, but I'd rather not have to maintain that fork and put the changes back into swag itself.

I am opening this PR as a more of a place for discussion for which parts can be merged into swag.

### DIFF
--- a/gen/gen.go
+++ b/gen/gen.go
@@ -40,6 +40,18 @@ func New() *Gen {
 	}
 }
 
+type GenConfig struct {
+	// OutputDir represents the output directory for all the generated files
+	OutputDir string
+
+	// InstanceName is used to get distinct names for different swagger documents in the
+	// same project. The default value is "swagger".
+	InstanceName string
+
+	// GeneratedTime whether swag should generate the timestamp at the top of docs.go
+	GeneratedTime bool
+}
+
 // Config presents Gen configurations.
 type Config struct {
 	// SearchDir the swag would be parse,comma separated if multiple
@@ -91,10 +103,6 @@ type Config struct {
 
 // Build builds swagger json file  for given searchDir and mainAPIFile. Returns json
 func (g *Gen) Build(config *Config) error {
-	if config.InstanceName == "" {
-		config.InstanceName = swag.Name
-	}
-
 	searchDirs := strings.Split(config.SearchDir, ",")
 	for _, searchDir := range searchDirs {
 		if _, err := os.Stat(searchDir); os.IsNotExist(err) {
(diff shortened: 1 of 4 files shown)


gen/gen.go
```diff
@@ -135,7 +143,23 @@ func (g *Gen) Build(config *Config) error {
...
+
+	return g.Generate(p.GetSwagger(), &GenConfig{
+		OutputDir:     config.OutputDir,
+		InstanceName:  config.InstanceName,
+		GeneratedTime: config.GeneratedTime,
+	})
+}
+
+// Generate outputs a swagger spec
+func (g *Gen) Generate(swagger *spec.Swagger, config *GenConfig) error {
```
e-nikolov, line 155: I moved out of the Build function the parts that deal with generating the files that store the generated swagger spec, because in my case I don't generate the specs from the swag binary, but I want to create the same files that swag does.

operation.go
```diff
@@ -810,6 +810,10 @@ func (operation *Operation) parseCombinedObjectSchema(refType string, astFile *a
 	}), nil
 }
 
+func (operation *Operation) ParseAPIObjectSchema(schemaType, refType string, astFile *ast.File) (*spec.Schema, error) {
```
e-nikolov, line 813: I need to use this function for creating a swagger schema for a type name

parser.go
```diff
@@ -1281,21 +1281,55 @@ func defineTypeOfExample(schemaType, arrayType, exampleValue string) (interface{
 	return nil, fmt.Errorf("%s is unsupported type in example value %s", schemaType, exampleValue)
 }
 
-// GetAllGoFileInfo gets all Go source files information for given searchDir.
-func (parser *Parser) getAllGoFileInfo(packageDir, searchDir string) error {
-	return filepath.Walk(searchDir, func(path string, f os.FileInfo, _ error) error {
-		if err := parser.Skip(path, f); err != nil {
+// GetAllGoFileInfoAndParseTypes gets all Go source files information for given searchDir and parses the types from them.
+func (parser *Parser) GetAllGoFileInfoAndParseTypes(searchDir string) error {
```
e-nikolov, line 1285: I need this in order to be able to go through packages one by one and incrementally parse the types from them.

I saw that ParseAPIMultiSearchDir() first collects all types from the astFiles and then parses them all at once via parser.packages.ParseTypes(), but in my case that would mean having to do a double pass through all routes, which I'd rather not have to do and instead have a function that parses the types of a single package.

ubogdan:
@e-nikolov Thanks for contributing to this project. The idea of opening a pull request for your project requirements without discussing it on an issue it's very uninspired.

e-nikolov:
I'm sorry if a PR wasn't your preference. I thought it would benefit the discussion to make the proposed changes more visible. I'm not sure if it would be as easy to talk about each line in an issue.

ubogdan:
@e-nikolov I'm afraid these changes may introduce some breaking changes, and I can't take this responsibility.

e-nikolov:
I understand not wanting to break backwards compatibility. Do you think it would be possible to export

e-nikolov:
I am mainly referring to exporting:
- [operation.parseAPIObjectSchema()](https://github.com/swaggo/swag/blob/a8ef569b67b5614c07f27fcb5a5757b9a948f979/operation.go#L813)
- [parser.packages](https://github.com/swaggo/swag/blob/32f02b99260db21a5bf0c28757a89e015682c574/parser.go#L88)
- [PackagesDefinitions.parseTypesFromFile()](https://github.com/swaggo/swag/blob/f631188105ab9e3557214884a9c2e81f7913a2f6/packages.go#L107)
- [the parts of gen.Build() that output the directories based on a provided swagger spec](https://github.com/swaggo/swag/blob/master/gen/gen.go#L113)

PR #1107, opened 2022-01-20
fix: split extensions param respecting open/close brackets

helder-jaspion:
**Describe the PR**
split extensions param respecting open/close brackets
`// @Param range query string true "Range" default([0, 9]) extensions(x-example=[0,9])`

**Relation issue**
None.

**Additional context**
When using extensions param with text containing comma the extensions were getting splited wrong.

### DIFF
--- a/field_parser.go
+++ b/field_parser.go
@@ -207,7 +207,9 @@ func (ps *tagBaseFieldParser) ComplementSchema(schema *spec.Schema) error {
 	extensionsTag := ps.tag.Get(extensionsTag)
 	if extensionsTag != "" {
 		structField.extensions = map[string]interface{}{}
+		extensionsTag := strings.ReplaceAll(extensionsTag, `,,`, `{{escape-comma}}`)
 		for _, val := range strings.Split(extensionsTag, ",") {
+			val = strings.ReplaceAll(val, `{{escape-comma}}`, ",")
 			parts := strings.SplitN(val, "=", 2)
 			if len(parts) == 2 {
 				structField.extensions[parts[0]] = parts[1]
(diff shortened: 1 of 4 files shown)


ubogdan:
It this PR related to any open issue?

helder-jaspion:
No. I needed to use extensions params and had problems with texts containing comma and then opened this PR.

helder-jaspion:
@ubogdan could you please approve running the workflows?

ubogdan:
> @ubogdan could you please approve running the workflows?

It seems I missed this notification.

field_parser.go
```diff
@@ -207,7 +207,9 @@ func (ps *tagBaseFieldParser) ComplementSchema(schema *spec.Schema) error {
 	extensionsTag := ps.tag.Get(extensionsTag)
 	if extensionsTag != "" {
 		structField.extensions = map[string]interface{}{}
+		extensionsTag := strings.ReplaceAll(extensionsTag, `,,`, `{{escape-comma}}`)
 		for _, val := range strings.Split(extensionsTag, ",") {
```
ubogdan, line 211: Adding double-quoted is an ugly hack. 

The solution is to line instead of split by comma by respection open/close brackets.

ubogdan (changes requested):

field_parser.go
helder-jaspion: I agree. double-comma was a quick fix. I'll try the solution you suggested

helder-jaspion: Changed as suggested

ubogdan (approved):
LGTM

ubogdan:
@helder-jaspion Thanks for your contribution!

Issue #1032, opened 2021-10-21
Slow performance when generating spec with external dependencies


matrixik:
**Describe the bug**
When using external dependencies `swag` is becoming really slow when generating swagger specification files.
I created example repository with simple code: https://github.com/matrixik/swag-perf
On my computer with CPU Intel Core i7-4790 it's taking around 30 seconds to generate specification files. This is simple example on stationary computer. We have some internal projects that on laptops take over 10 minutes to generate spec files.

At the same time same code will generate swagger spec files using https://github.com/go-swagger/go-swagger in around 3 seconds. 
We changed some smaller internal projects to `go-swagger` and it's still generate everything in under 5 seconds when `swag` need over 1 minute and 30 seconds to output spec files on my computer. Output from both is almost the same but `swag` is minimum 10 times slower.
We would like to stay with `swag` because it's annotations are more readable but long generation times are getting in the way.

**To Reproduce**

*Clone example repository*
`git clone https://github.com/matrixik/swag-perf.git`

*Clone and fix swag*

`git clone https://github.com/swaggo/swag.git && cd swag`

`git checkout v1.7.3`

`git apply ../swag-perf/swag_annotation_error.patch` - needed to avoid exiting on wrong annotations in dependencies (or you can use v1.6.7 without patching)
```diff
diff --git i/parser.go w/parser.go
index aaa8112..fc6801e 100644
--- i/parser.go
+++ w/parser.go
@@ -636,7 +636,8 @@ func (parser *Parser) ParseRouterAPIInfo(fileName string, astFile *ast.File) err
 			for _, comment := range astDeclaration.Doc.List {
 				err := operation.ParseComment(comment.Text, astFile)
 				if err != nil {
-					return fmt.Errorf("ParseComment error in file %s :%+v", fileName, err)
+					fmt.Println("Error found in file:", fileName, "error:", err)
+					// return fmt.Errorf("ParseComment error in file %s :%+v", fileName, err)
 				}
 			}
 ```

`go build -o swag_fix ./cmd/swag`

*Build swagger spec*

`cd ../swag-perf`

Generate Swagger spec file:

With https://github.com/swaggo/swag

* v1.7.3:
  `time ../swag/swag_fix init -o swagger/docs --md swagger/markdown --parseDependency true`
* v1.6.7:
    `time swag init -o swagger/docs --md swagger/markdown`

With https://github.com/go-swagger/go-swagger

[...]

PR #1108, opened 2022-01-21
Imporve performance when generating spec with external dependencies

pytimer:
**Describe the PR**

Directly use `go list` command instead of `depth`, because `depth` use go standard library `build.Import`  and call the `build.Import` too many times, it will spent too many times when the large dependencies.

Because `swag` only needs to know which files are dependent, and does not need to get the level of dependencies, all dependencies can be directly through `go list` command, and `swag` not need to go through deep loop again, but if use `depth`, `swag` still need to loop too many times, so i refactor the swag get ast files code.

**Relation issue**

Fixes: https://github.com/swaggo/swag/issues/1032

**Additional context**

Added the `--parseGoList` flag, the default value is true default. If someone does not want to enable it, it can be set to false when init, if it's set to false, `swag` still use `depth` for dependency parser.

### DIFF
--- a/golist.go
+++ b/golist.go
@@ -11,8 +11,7 @@ import (
 )
 
 func listPackages(ctx context.Context, dir string, env []string, args ...string) (pkgs []*build.Package, finalErr error) {
-	goArgs := append([]string{"list", "-json", "-e"}, args...)
-	cmd := exec.CommandContext(ctx, "go", goArgs...)
+	cmd := exec.CommandContext(ctx, "go", append([]string{"list", "-json", "-e"}, args...)...)
 	cmd.Env = env
 	cmd.Dir = dir
 
@@ -28,18 +27,21 @@ func listPackages(ctx context.Context, dir string, env []string, args ...string)
 		}
 	}()
 
-	if err := cmd.Start(); err != nil {
+	err = cmd.Start()
+	if err != nil {
 		return nil, err
 	}
 	dec := json.NewDecoder(stdout)
 	for dec.More() {
 		var pkg build.Package
-		if err := dec.Decode(&pkg); err != nil {
+		err = dec.Decode(&pkg)
+		if err != nil {
 			return nil, err
 		}
 		pkgs = append(pkgs, &pkg)
 	}
-	if err := cmd.Wait(); err != nil {
+	err = cmd.Wait()
+	if err != nil {
 		return nil, err
 	}
 	return pkgs, nil
@@ -57,16 +59,17 @@ func (parser *Parser) getAllGoFileInfoFromDepsByList(pkg *build.Package) error {
 	}
 
 	srcDir := pkg.Dir
+	var err error
 	for i := range pkg.GoFiles {
-		path := filepath.Join(srcDir, pkg.GoFiles[i])
-		if err := parser.parseFile(pkg.ImportPath, path, nil); err != nil {
+		err = parser.parseFile(pkg.ImportPath, filepath.Join(srcDir, pkg.GoFiles[i]), nil)
+		if err != nil {
 			return err
 		}
 	}
 
 	for i := range pkg.CFiles {
-		path := filepath.Join(srcDir, pkg.CFiles[i])
-		if err := parser.parseFile(pkg.ImportPath, path, nil); err != nil {
+		err = parser.parseFile(pkg.ImportPath, filepath.Join(srcDir, pkg.CFiles[i]), nil)
+		if err != nil {
 			return err
 		}
 	}
--- a/golist.go
+++ b/golist.go
@@ -53,11 +53,6 @@ func (parser *Parser) getAllGoFileInfoFromDepsByList(pkg *build.Package) error {
 		return nil
 	}
 
-	// Skip cgo
-	if pkg.Name == "C" {
-		return nil
-	}
-
 	srcDir := pkg.Dir
 	var err error
 	for i := range pkg.GoFiles {
@@ -67,8 +62,9 @@ func (parser *Parser) getAllGoFileInfoFromDepsByList(pkg *build.Package) error {
 		}
 	}
 
-	for i := range pkg.CFiles {
-		err = parser.parseFile(pkg.ImportPath, filepath.Join(srcDir, pkg.CFiles[i]), nil)
+	// parse .go source files that import "C"
+	for i := range pkg.CgoFiles {
+		err = parser.parseFile(pkg.ImportPath, filepath.Join(srcDir, pkg.CgoFiles[i]), nil)
 		if err != nil {
 			return err
 		}
--- a/golist.go
+++ b/golist.go
@@ -1,0 +1,75 @@
+package swag
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"go/build"
+	"os/exec"
+	"path/filepath"
+)
+
+func listPackages(ctx context.Context, dir string, env []string, args ...string) (pkgs []*build.Package, finalErr error) {
+	goArgs := append([]string{"list", "-json", "-e"}, args...)
+	cmd := exec.CommandContext(ctx, "go", goArgs...)
+	cmd.Env = env
+	cmd.Dir = dir
+
+	stdout, err := cmd.StdoutPipe()
+	if err != nil {
+		return nil, err
+	}
+	var stderrBuf bytes.Buffer
+	cmd.Stderr = &stderrBuf
+	defer func() {
+		if stderrBuf.Len() > 0 {
+			finalErr = fmt.Errorf("%v\n%s", finalErr, stderrBuf.Bytes())
+		}
+	}()
+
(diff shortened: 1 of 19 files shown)


ubogdan:
Unfortunately, I will have a busy week.

@easonlin404 can you please assist with a CR here?

magandrez:
@ubogdan @easonlin404 Is there any way to contribute to this change? please, let know.

ubogdan:
This implementation introduces a dependency over an external binary, the go binary, that comes with some new challenges :
1. The output format may change over time.
2. You may get slightly different output on different go versions.
3. You may run the generator in an environment where the GO binary it's not in PATH.
4. You may run the generator in an environment where the GO binary is not the same as your one to build your project.

I prefer to see this done programmatically by importing a go library that handlers go.mod and calling the specific methods.

The solution proposed by @pytimer may look like a quick fix at this time. Once the feature gets merged to master, the community will start using it, more issues will arrive, and code complexity will grow.

pytimer:
@ubogdan `swag` currently already depends on the go binary, now if the env has not go binary, `swag init` failed, because init first call `go list` to get pkg name, the code is [getPkgName](https://github.com/swaggo/swag/blob/master/parser.go#L330).
`depth` call `build.Default`, this code also call the go binary file, if no go binary, it will return `ErrRootPkgNotResolved`, this pr does not newly introduce the go binary.

However, the two points 2 and 4 above @ubogdan said do need to be pay attention to. I don't know much about the output of multiple versions of go list. I only tested it on versions of go 1.16 and 1.17. 
I will try to understand the output format of go list in different versions. If anyone else is familiar with it, please point it out.

ubogdan:
> @ubogdan swag currently already depends on the go binary, 
Please don't return this as an excuse. 
I explained why I wouldn't say I like the idea, but didn't refuse the implementation. 
It's more simple for me to Approve and Merge and after that, I can tag you in every issue where a community member is complaining something is wrong.

g1eny0ung:
Any updates? After my local testing, this will greatly improve the generation speed.

ubogdan:
I will consider this total silence as an aproval. 
@pytimer I will go forward with a code review in the next few days, please confirm you are still interested to finish this PR.

pytimer:
Thanks.
I read some docs about not use binary execute command to do it recently, but unfortunately i don't find.

I try golang.org/x/tools/go/packages.Load, but it not works well.

Perhaps use `go list` now and find the better ways to optimize in later version, what's your recommend?

ubogdan:
We will go with the current implementation. Please confirm you still intend to finish this PR so I can go with a CR.

pytimer:
> We will go with the current implementation. Please confirm you still intend to finish this PR so I can go with a CR. 

Yes，please tell me if need I do.

golist.go
```diff
@@ -0,0 +1,75 @@
...
+	"fmt"
+	"go/build"
+	"os/exec"
+	"path/filepath"
+)
+
+func listPackages(ctx context.Context, dir string, env []string, args ...string) (pkgs []*build.Package, finalErr error) {
+	goArgs := append([]string{"list", "-json", "-e"}, args...)
+	cmd := exec.CommandContext(ctx, "go", goArgs...)
+	cmd.Env = env
```
ubogdan, line 16: `goArgs` is used only once
```go
	cmd := exec.CommandContext(ctx, "go", append([]string{"list", "-json", "-e"}, args...)...)
```

golist.go
```diff
@@ -0,0 +1,75 @@
...
+	}()
+
+	if err := cmd.Start(); err != nil {
+		return nil, err
+	}
+	dec := json.NewDecoder(stdout)
+	for dec.More() {
+		var pkg build.Package
+		if err := dec.Decode(&pkg); err != nil {
+			return nil, err
```
ubogdan, line 38: Same here ...
```go
                err = dec.Decode(&pkg);
		if  err != nil {
```

golist.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+		pkgs = append(pkgs, &pkg)
+	}
+	if err := cmd.Wait(); err != nil {
+		return nil, err
```
ubogdan, line 43: Same here
```go
        err = cmd.Wait(); 
	if err != nil {
		return nil, err
```

golist.go
```diff
@@ -0,0 +1,75 @@
...
+	if ignoreInternal { // ignored internal
+		return nil
+	}
+
+	// Skip cgo
+	if pkg.Name == "C" {
+		return nil
+	}
+
+	srcDir := pkg.Dir
```
ubogdan, line 59: var err error

golist.go
```diff
@@ -0,0 +1,75 @@
...
+		path := filepath.Join(srcDir, pkg.GoFiles[i])
+		if err := parser.parseFile(pkg.ImportPath, path, nil); err != nil {
+			return err
+		}
+	}
+
+	for i := range pkg.CFiles {
+		path := filepath.Join(srcDir, pkg.CFiles[i])
+		if err := parser.parseFile(pkg.ImportPath, path, nil); err != nil {
+			return err
```
ubogdan, line 70: Same here
```go
                err = parser.parseFile(pkg.ImportPath, filepath.Join(srcDir, pkg.CFiles[i]), nil); 
		if err != nil {
			return err

File: parser.go [hunk 6]
```diff
@@ -140,6 +141,9 @@ type Parser struct {
 
 	// Overrides allows global replacements of types. A blank replacement will be skipped.
 	Overrides map[string]string
+
+	// ParseGoList whether swag use go list to parse dependency
+	ParseGoList bool
```
ubogdan, line 146: Make this private, will be updated by 

```go
// ParseUsingGoList --fill--compent---please.
func ParseUsingGoList(Enabled bool) func(*Parser) {
	return func(p *Parser) {
		p.parseGoList = Enabled
	}
}
```

ubogdan (changes requested):
See comments ... Also include the usage example in the README.md

pytimer:
@ubogdan Update the code, you can review it if you have time, thanks.

ubogdan (approved):
LGTM

ubogdan:
> @ubogdan Update the code, you can review it if you have time, thanks.
> 
> Should i squash all commits after review or now?

We are using  "Squash and Merge" for all the contributions.

golist.go
```diff
@@ -0,0 +1,78 @@
...
+func (parser *Parser) getAllGoFileInfoFromDepsByList(pkg *build.Package) error {
+	ignoreInternal := pkg.Goroot && !parser.ParseInternal
+	if ignoreInternal { // ignored internal
+		return nil
+	}
+
+	// Skip cgo
+	if pkg.Name == "C" {
+		return nil
+	}
```
ubogdan, line 59: Please cover this by calling the function with a build.Package instance.

ubogdan:
Please improve code coverage so we can merge it.

golist.go
pytimer: ok, i will add some testing on this Saturday, i don't have the bandwidth these days, sorry.

akhmadKurniawan (commented):
ok

sdghchj:
> I expect to see at least one opinion from @easonlin404 or @sdghchj regarding this matter before going forward to do a real CR and approve it.

As far as I see, `go list` will find out all its dependent packages, I am not sure whether it's more efficient in a huge project or not.

magandrez:
Any ETA to land this? @ubogdan

ubogdan:
@magandrez, I'm waiting for @pytimer to improve code coverage so we can merge it.

g1eny0ung:
@ubogdan I wonder if it is possible to draft a new release after merging this PR? I've been looking forward to this improvement for a long time. Very grateful to everyone for this.

pytimer:
@ubogdan How can i rerun github Actions and codecov?

ubogdan:
Good job. Please resolve conflicts.

pytimer:
@ubogdan solve conflicts, please review it, thanks.

ubogdan:
Looks like code coverage is still failing :(

pytimer:
Maybe i merge master branch code, add test code in the latest commit.

pytimer:
Please rerun test codecov, thanks.

pytimer:
@ubogdan Can you tell me how to fix codecov/patch, i don't understand where the problem, thanks.

ubogdan:
Maybe you need to write unit tests this parts of the code: https://github.com/swaggo/swag/pull/1108/checks?check_run_id=6346809522?

ubogdan (approved):
LGTM

ubogdan:
> y ETA regarding the required changes

We will try to include this in the next release. I was hoping we would get better code coverage.

g1eny0ung:
@ubogdan Can we release a version? It's been a long time since this PR was merged. 😢

ubogdan:
released in v1.8.3

g1eny0ung:
> released in v1.8.3

Many thanks!

Issue #1113, opened 2022-01-28
Replacing init in doc.go as a Register method


Ferrany1:
**Is your feature request related to a problem? Please describe.**
As for it would be easier to register swagger by myself through a direct import

**Describe the solution you'd like**
Replacing init with Register method

**Describe alternatives you've considered**
Adding a configuration (by default keep init) else self register

ubogdan:
@Ferrany1 that's a great idea but we can't break compatibility. 

This may be a feature to add maybe in v2.x.x.

Ferrany1:
Cant it be configured thorough a cli for now? (to not break current usages?)

ubogdan:
I'm looking at the pr you make and I'm not sure what exactly do you want to accomplish. 

I've understood the fact you may like to have 2 swagger docs in the same packages. Still not sure what is wrong with init func () ?

Ferrany1:
@ubogdan No, I want to have 1 doc file and make a direct call for register to handle error and get doc name to handle in ReadDoc later while using my middleware.
As for me init func is worse, because I can't control it, also direct import > blank import as injection

ubogdan:
> @ubogdan No, I want to have 1 doc file and make a direct call for register to handle error and get doc name to handle in ReadDoc later while using my middleware.

So then why are u naming the files by instance name ?

Ferrany1:
Because in my case I have different microservices stored in monorepo and all documentation stored in doc folder, so in this case each doc will overwrite previous, that's why instance name may be necessary in monorepo cases

ubogdan:
The init method allows you to register multiple instance names. Why can't you use that?

Ferrany1:
I want to get instance name as a string as well to pass to handler for ReadDoc method

ubogdan:
Okay, but we can introduce breaking changes. 

As you may see this repo has 5k followers, you can imagine there are over 5k projects depending on it.

Ferrany1:
But right now I'm not proposing to break contract, just to make option to make a modification

Ferrany1:
It's up to user now handle error or not, this won't break any current repos

ubogdan:
If you look closer, the same pattern was used by the SQL drivers. See MySQL implementation func init() {
	sql.Register("mysql", &MySQLDriver{})
}

That's been said, the Register method signature cannot be changed and the init method will not be removed. 

Please figure out a different way to implement this. 

[...]

PR #1115, opened 2022-01-28
Register func to doc file option addition

Ferrany1:
**Describe the PR**
Register func to doc file option addition

**Relation issue**
Replacing init in doc.go as a Register method
https://github.com/swaggo/swag/issues/1113


Ferrany1:
@ubogdan Do I need to cover this case?

ubogdan:
Yes, please.

Ferrany1:
@ubogdan done

swagger.go
```diff
@@ -19,22 +19,28 @@ type Swagger interface {
...
 }
 
+// Documentation is an interface to register swagger document.
+type Documentation interface {
+	Register() (string, error)
+}
+
 // Register registers swagger for given name.
-func Register(name string, swagger Swagger) {
+func Register(name string, swagger Swagger) error {
```
ubogdan, line 28: Changing the function signature introduces a breaking change.

gen/gen_test.go
```diff
@@ -101,13 +101,17 @@ func TestGen_BuildInstanceName(t *testing.T) {
...
 	config.InstanceName = "custom"
+	config.DocUserRegister = true
 	goSourceFile = filepath.Join(config.OutputDir, config.InstanceName+"_"+"docs.go")
 	assert.NoError(t, New().Build(config))
 	expectedCode, err = ioutil.ReadFile(goSourceFile)
 	if err != nil {
 		require.NoError(t, err)
 	}
-	if !strings.Contains(string(expectedCode), "swag.Register(\"custom\", &s{})") {
+	println(string(expectedCode))
```
ubogdan, line 111: this has to be cleaned

ubogdan (changes requested):

swagger.go
Ferrany1: Isn't it better to allow user to handle error by himself instead of panicking?

PR #1116, opened 2022-01-30
Feature: Instance name func to doc file

Ferrany1:
**Describe the PR**
Feature: Instance name func to doc file

**Relation issue**
https://github.com/swaggo/swag/issues/1113


gen/gen.go
```diff
@@ -414,31 +414,37 @@ import (
...
+	instanceName: {{ printf "%q" .InstanceName }},
 }
 
-type s struct{}
+type s_{{ .InstanceName }} struct{}
 
-func (s *s) ReadDoc() string {
-	sInfo := SwaggerInfo
+func (s *s_{{ .InstanceName }}) ReadDoc() string {
+	sInfo := SwaggerInfo_{{ .InstanceName }}
```
ubogdan, line 447: Why do we need an additional s_InstanceName struct when we can define the ReadDoc() for swaggerInfo_InstanceName?
```
func (i *swaggerInfo_{{ .InstanceName }}) ReadDoc() string { 
```

ubogdan (changes requested):

gen/gen.go
Ferrany1: How are looking on defining swaggerInfo struct inside swag as exported struct and generating only var SwaggerInfo_InstanceName with all methods?

ubogdan: replace
```
func (s *s_{{ .InstanceName }}) ReadDoc() string {
```

with 
```
func (s *swaggerInfo_{{ .InstanceName }}) ReadDoc() string { 
```

remove
```
type s_{{ .InstanceName }} struct{}
sInfo := SwaggerInfo_{{ .InstanceName }}
```

update 
```
	if err := t.Execute(&tpl, sInfo); err != nil {
```
to 
```
	if err := t.Execute(&tpl, s); err != nil {
```

ubogdan: Does that make sense now ?

Ferrany1: How about doing it this way, it wont break any current cases

ubogdan: Removing ReadDoc() function will introduce breaking changes. 

See yourself by regenerating the documentation for example/basic and building the project with the proposed way.

Ferrany1: I've tried to, with importing local Swagger info struct, worked fine

Ferrany1: I'm referring to swag.SwaggerInfo, which matches Swagger interface

Ferrany1: @ubogdan what do you think about changes?

ubogdan: @Ferrany1 congratulations! moving the code from string constant to swag enable us to properly write unit tests against it. 
I'm quite caught with different things, but I will try to do a CR soon.

swagger_info.go
```diff
@@ -0,0 +1,60 @@
...
+	"strings"
+	"text/template"
+)
+
+// DocumentationReader is an interface to inject Swagger info reader.
+type DocumentationReader interface {
+	Swagger
+	InstanceName() string
+}
+
```
ubogdan, line 15: A good practice is to define the interface where you use it not where you implement it. Please remove this definition ...

gen/gen.go
```diff
@@ -405,68 +405,23 @@ var packageTemplate = `// Package {{.PackageName}} GENERATED BY THE COMMAND ABOV
...
-	Schemes     []string
-	Title       string
-	Description string
-}
-
-// SwaggerInfo holds exported Swagger Info so clients can modify it
-var SwaggerInfo = swaggerInfo{
+// SwaggerInfo_{{ .InstanceName }} holds exported Swagger Info so clients can modify it
+var SwaggerInfo_{{ .InstanceName }} = &swag.SwaggerInfo{
 	Version:     {{ printf "%q" .Version}},
```
ubogdan, line 414: Please rename the struct to Doc so the definition will not contain the swag string to times.
```
var SwaggerInfo_{{ .InstanceName }} = &swag.Spec {
```

swagger_info.go
```diff
@@ -0,0 +1,60 @@
...
+	Version          string
+	Host             string
+	BasePath         string
+	Schemes          []string
+	Title            string
+	Description      string
+	InfoInstanceName string
+	SwaggerTemplate  string
+}
+
```
ubogdan, line 27: Please rename the struct to Spec and also the filename spec.go

ubogdan (changes requested):
Good job till now. Would you mind making some other minor adjustments according to the comments?

swagger_info.go
Ferrany1: done

gen/gen.go
Ferrany1: Done

swagger_info.go
Ferrany1: done

ubogdan (approved):
LGTM

ubogdan (approved):
LGTM

Issue #1126, opened 2022-02-03
undefined: "github.com/swaggo/swag".Spec


bashdrew:
hi, i have been using this repo for a while now.  just a few hours ago, it suddenly threw out the error below:

docs/docs.go:11152:28: undefined: "github.com/swaggo/swag".Spec

may i know how to get rid of this error?  thank you.

momarques:
same for me here, it's been a while since I last used it. maybe it's something with v1.7.9?

bashdrew:
Yes, I switched back 1.7.8 and it worked.

ubogdan:
Please fill the report with swagger handler are u using and which version . 
example
  echo-swagger 1.0.0

We will investigate and will come back with a solution ASAP.

mehulgohil:
Yes getting the same error while initializing swagger in docker

I am using golang:1.16

![image](https://user-images.githubusercontent.com/28588619/152485215-72ef0d56-5a3d-458e-9c72-a3acef198287.png)

ubogdan:
@mehulgohil Thanks for letting us know, but you are not helping with so less details.

```shell
~ go get -u github.com/swaggo/swag/cmd/swag
go: downloading golang.org/x/tools v0.1.9
go: downloading github.com/urfave/cli v1.22.5
go: downloading github.com/go-openapi/swag v0.21.1
go: downloading github.com/cpuguy83/go-md2man/v2 v2.0.1
go: downloading github.com/cpuguy83/go-md2man v1.0.10
go: downloading github.com/russross/blackfriday/v2 v2.1.0
go: downloading github.com/mailru/easyjson v0.7.7
go: downloading golang.org/x/net v0.0.0-20220127200216-cd36cc0744dd
go: downloading github.com/russross/blackfriday v1.6.0
go: downloading golang.org/x/sys v0.0.0-20220128215802-99c3d69c2c27
go get: installing executables with 'go get' in module mode is deprecated.
	Use 'go install pkg@version' instead.
	For more information, see https://golang.org/doc/go-get-install-deprecation
	or run 'go help get' or 'go help install'.

~ swag -v
swag version v1.7.9

~ go version
go version go1.17.5 darwin/amd64
```

AxelDum:
I had the same issue I was installing swag and generating docs in docker with a go.mod not necessarily matching latest

In docker file i have:
```
FROM golang:1.17-alpine
RUN go mod download
RUN go install github.com/swaggo/swag/cmd/swag@latest
[...]

PR #1127, opened 2022-02-05
fix: renaming SwaggerInfo causes a breaking change

ubogdan:
**Describe the PR**
* fix breaking change introduced by the generation of multiple api documents in the same doc folder.

**Relation issue**
#1126

**Additional context**
None.


ubogdan:
@easonlin404  Not sure if we need to release 1.8.0 only with this patch or we should wait for other contributions.

ubogdan:
Alternatively, we can release v1.7.9p1 only with this patch ?

Issue #1136, opened 2022-02-13
Linter fails


pilinux:
**Linter fails**

```
docs/docs.go:7:7: don't use underscores in Go names; const docTemplate_swagger should be docTemplateSwagger
docs/docs.go:102:5: don't use underscores in Go names; var SwaggerInfo_swagger should be SwaggerInfoSwagger
```

**To Reproduce**

`swag init`

`golint ./...`

`revive ./...`

**Expected behavior**

The convention in Go is to use MixedCaps or mixedCaps (simply camelCase) rather than underscores to write multi-word names.

**Your swag version**
`1.7.9`

**Your go version**
`1.17.5`

**Desktop (please complete the following information):**
 - OS: [`Ubuntu`]
 - Browser: [`chrome`]

**Additional context**
`swag cli` should generate `docTemplateSwagger` instead of `docTemplate_swagger`
 and `SwaggerInfoSwagger` instead of `SwaggerInfo_swagger`

PR #1137, opened 2022-02-13
Fix: Linter fails (#1136)

pilinux:
**Describe the PR**
Fix: Linter fails (#1136)


swagger.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 // Name is a unique name be used to register swag instance.
-const Name = "swagger"
+const Name = "Swagger"
```
ubogdan, line 10: I don't think changing this value may be a good idea.

ubogdan (changes requested):
Please undo the changes regarding the default swagger instance name.

swagger.go
pilinux: `const Name = "swagger"` will generate `docTemplateswagger` and `SwaggerInfoswagger` which is OK from the perspective of a linter, but a bit difficult to read from a human's perspective. That's why I capitalized it. And this change also didn't break any other part of the full repo.

After reading this comment, if you still believe that it is a better idea to keep it as `"swagger"`, kindly let me know. I will revert it back.

Thank you.

ubogdan: The generates SwaggerInfo if Instance name == "swagger" and SwaggerInfo+InstanceName if not.
```
var SwaggerInfo{{ if ne .InstanceName "Swagger" }}{{ .InstanceName }} {{- end }} = &swag.Spec{
```

Please roll back that change.

ubogdan:
Please update the doc of all examples with the latest "genetator version"

swagger.go
pilinux: done.

gen/gen.go
```diff
@@ -407,21 +407,21 @@ package {{.PackageName}}
 
 import "github.com/swaggo/swag"
 
-const docTemplate_{{ .InstanceName }} = ` + "`{{ printDoc .Doc}}`" + `
+const docTemplate{{ .InstanceName }} = ` + "`{{ printDoc .Doc}}`" + `
```
ubogdan, line 411: Ok then let's update this to 
```
const docTemplate{{ if ne .InstanceName "swagger" }}{{ .InstanceName }} {{- end }} = ` + "`{{ printDoc .Doc}}`" + `
```

gen/gen.go
```diff
@@ -407,21 +407,21 @@ package {{.PackageName}}
...
+var SwaggerInfo{{ if ne .InstanceName "swagger" }}{{ .InstanceName }} {{- end }} = &swag.Spec{
 	Version:     {{ printf "%q" .Version}},
 	Host:        {{ printf "%q" .Host}},
 	BasePath:    {{ printf "%q" .BasePath}},
 	Schemes:     []string{ {{ range $index, $schema := .Schemes}}{{if gt $index 0}},{{end}}{{printf "%q" $schema}}{{end}} },
 	Title:       {{ printf "%q" .Title}},
 	Description: {{ printf "%q" .Description}},
 	InfoInstanceName: {{ printf "%q" .InstanceName }},
-	SwaggerTemplate: docTemplate_{{ .InstanceName }},
+	SwaggerTemplate: docTemplate{{ .InstanceName }},
```
ubogdan, line 421: and this to 
```
	SwaggerTemplate: docTemplate{{ if ne .InstanceName "swagger" }}{{ .InstanceName }} {{- end }}
```

ubogdan (changes requested):
I see what you mean now. Please update according to the changes and regenerate docs once again.
Thanks for your patience.

ubogdan:
@pilinux it looks I'm not in very good shape this week. So please bear with me.

gen/gen.go
pilinux: Done.

gen/gen.go
pilinux: Done.

ubogdan (approved):
LGTM

ubogdan:
@pilinux Thanks for your contribution.

Issue #1140, opened 2022-02-16
String Example Value


mschneider82:
Hi,

I tried the new example feature for string:
`
// @Param email body string true "message/rfc822" example(Subject: Testmail\r\n\r\nBody Message\r\n)
`
Going to the swagger page it still shows "string" as the default Example Value. Howto change this to the example value? I would like to add a basic test mail as the text (which has a subject header and a message body)

![image](https://user-images.githubusercontent.com/8426497/154281520-29b73bfd-220b-4765-9408-b74968db7d34.png)

ubogdan:
Not sure why do you want to implement rfc822 over HTTP but you should try the following

```go
type Message struct {
    Subject string `example:"Test mail"`
    Body string `example:"Message"`
}
// @consume json
// @Param email body Message true "Json Payload"
```

mschneider82:
A Content-Type of "message/rfc822" indicates that the body contains an encapsulated message, with the syntax of an RFC 822 message. Sending json ist wrong here

A valid Body could be the String:
Subject: Testmail\r\n\r\nBody Message\r\n

It has one Header followed by a newline that indicates that the Body Starts with the Text "Body messeage"

ubogdan:
See OAS 2 specs regarding [parameter object](https://spec.openapis.org/oas/v2.0#parameter-object). 

You are free to use "message/rfc822" or anything else that's not in the specs, but you are on your own.

mschneider82:
I have tried your Message struct, it gives an json record in the "edit Value" field instead of an email.

So its not possible to enter an normal string as body example?

mschneider82:
I got what i want by editing the docs/docs.go file:

```
                     {
-                        "example": "\"Subject: Testmail\\r\\n\\r\\nBody Message\\r\\n\"",
                         "description": "message/rfc822",
                         "name": "email",
                         "in": "body",
                         "required": true,
                         "schema": {
-                            "type": "string"
+                            "type": "string",
+                            "example": "Subject: Testmail\r\n\r\nBody Message\r\n"
                         }
                     }
                     
```

mschneider82:
results in:
![image](https://user-images.githubusercontent.com/8426497/155489242-a4b2a8c3-0ef6-4908-8935-09475fb96011.png)

mschneider82:
@ubogdan  what do you think about the PR ? Other idea is to include the Schema Example as default when body and example is used

davidkhala:
[...]

PR #1156, opened 2022-03-08
feat: SchemaExample: allow \r \n \t

mschneider82:
#1140

### DIFF
--- a/operation.go
+++ b/operation.go
@@ -535,6 +535,9 @@ func setDefault(param *spec.Parameter, schemaType string, value string) error {
 	return nil
 }
 
+// controlCharReplacer replaces \r \n \t in example string values
+var controlCharReplacer = strings.NewReplacer(`\r`, "\r", `\n`, "\n", `\t`, "\t")
+
 func setSchemaExample(param *spec.Parameter, schemaType string, value string) error {
 	val, err := defineType(schemaType, value)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


operation.go
```diff
@@ -544,7 +555,8 @@ func setSchemaExample(param *spec.Parameter, schemaType string, value string) er
 	if param.Schema == nil {
 		return nil
 	}
-	param.Schema.Example = val
+
+	param.Schema.Example = unescapeExample(val)
```
ubogdan, line 559: please move the unescapeExample() code here.

operation.go
```diff
@@ -553,7 +565,7 @@ func setExample(param *spec.Parameter, schemaType string, value string) error {
 	if err != nil {
 		return nil // Don't set a example value if it's not valid
 	}
-	param.Example = val
+	param.Example = unescapeExample(val)
 	return nil
```
ubogdan, line 569: Why are u making this change here?

ubogdan (changes requested):
See comments...

operation.go
mschneider82: To also Support this the replacing in Example() , thats why i put it into a extra func

mschneider82: any advice? remove or keep it for Example?

ubogdan: Please remove the usage on Example. It may introduce a breaking change not covered by unit tests.

mschneider82: done

ubogdan (dismissed):
LGTM

ubogdan:
Can you please update the README.md accordingly?

ubogdan (approved):
LGTM

ubogdan:
@mschneider82 Thanks for your contribution.

PR #1158, opened 2022-03-10
Feature/security or

Lehp:
- remove unnecessary comment,
- add or-functionality to parser_test

### DIFF
--- a/README.md
+++ b/README.md
@@ -813,6 +813,14 @@ Make it AND condition
 // @Security OAuth2Application[write, admin]
 ```
 
+Make it OR condition
+
+```go
+// @Security ApiKeyAuth || firebase
+// @Security OAuth2Application[write, admin] || APIKeyAuth
+```
+
+
 ### Add a description for enum items
 
 ```go
--- a/operation.go
+++ b/operation.go
@@ -629,30 +629,38 @@ func (operation *Operation) ParseRouterComment(commentLine string) error {
 	return nil
 }
 
+type SecurityMap map[string][]string
+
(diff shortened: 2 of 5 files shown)


Lehp:
@ubogdan I added or to the general check.

How can i use this functionality locally? I tried using the master branch and i saw my code in the vendor-folder, but it didn't work.

ubogdan (approved):
LGTM

Lehp:
@ubogdan Hey ubogdan I've been looking atleast 5h for the reason my code doesn't work. Everything executes fine. Every tests is ok. But in the End the file does not contain the object parseSecurityComment returns. Could you give me a hint? I don't know where to search anymore.

The input comment seen in a Test:
image

The result of ParseSecurityComment: 
image

Left -> comment in goFile right-> result in swagJson
image

Lehp:
> 

Hey ubogdan sorry for the late response. The function works exactly as expected. It adds both values into the array which creates a security option with two options so to say. I do not understand where this gets written into a file because thats where the error must happen.

Lehp:
> > 
> 
> Hey ubogdan sorry for the late response. The function works exactly as expected. It adds both values into the array which creates a security option with two options so to say. I do not understand where this gets written into a file because thats where the error must happen.
> 
> This test works fine and it's exactly what i need in my swaggerfile image

I made some more testing and found out that comments in Tests must be read correctly cuz following test ran succeeded. I really don't know it doesn't look like my code gets executed at all. Does the init command of the cli somehow use an older version of swag?


image

PR #1174, opened 2022-04-12
added security definition description

pmorelli92:
**Describe the PR**

Added description capabilities for the security definitions types.

**Additional context**

I know this field is not mandatory, but another tool I am working with, relies on me adding the descriptions, otherwise it will break. I would need pretty please a tag if this gets merged 🙏


ubogdan:
@pmorelli92 I perfectly understand the use case, please open a feature request https://github.com/swaggo/swag/issues examining this implementation requirement and I will follow you with a CR.

pmorelli92:
Done @ubogdan https://github.com/swaggo/swag/issues/1175

README.md
```diff
@@ -444,11 +444,11 @@ Besides that, `swag` also accepts aliases for some MIME Types as follows:
...
+| securitydefinitions.oauth2.application  | [OAuth2 application](https://swagger.io/docs/specification/authentication/oauth2/) auth.       | tokenUrl, scope, sec.def.description                   | // @securitydefinitions.oauth2.application OAuth2Application |
+| securitydefinitions.oauth2.implicit     | [OAuth2 implicit](https://swagger.io/docs/specification/authentication/oauth2/) auth.          | authorizationUrl, scope, sec.def.description           | // @securitydefinitions.oauth2.implicit OAuth2Implicit       |
+| securitydefinitions.oauth2.password     | [OAuth2 password](https://swagger.io/docs/specification/authentication/oauth2/) auth.          | tokenUrl, scope, sec.def.description                   | // @securitydefinitions.oauth2.password OAuth2Password       |
+| securitydefinitions.oauth2.accessCode   | [OAuth2 access code](https://swagger.io/docs/specification/authentication/oauth2/) auth.       | tokenUrl, authorizationUrl, scope, sec.def.description | // @securitydefinitions.oauth2.accessCode OAuth2AccessCode   |
```
ubogdan, line 451: Please update sec.def.description with "description" for the parameter column and add also the "description" feature in the example column.

ubogdan (changes requested):
Well done. 
We need to have a clear readme so our users will understand how to use it.

parser.go
```diff
@@ -472,35 +472,45 @@ func parseGeneralAPIInfo(parser *Parser, comments []string) error {
...
-			attrMap, _, _, err := parseSecAttr(attribute, []string{"@in", "@name"}, comments[i+1:])
+			attrMap, _, extensions, err := parseSecAttr(attribute, []string{"@in", "@name"}, comments[i+1:])
 			if err != nil {
 				return err
 			}
-			parser.swagger.SecurityDefinitions[value] = spec.APIKeyAuth(attrMap["@name"], attrMap["@in"])
+			secDef := spec.APIKeyAuth(attrMap["@name"], attrMap["@in"])
+			tryAddDescription(secDef, extensions)
+			parser.swagger.SecurityDefinitions[value] = secDef
 		case "@securitydefinitions.oauth2.application":
```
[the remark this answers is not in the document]
pmorelli92: Solved :)

README.md
pmorelli92: I added the example on the table below where the `parameter annotation | example` lives as I cannot do line breaks on the example cell of the mentioned table. Hope that is ok.

pmorelli92: @ubogdan , as `sec.def.description` is a "line" parameter such as `header`, `in`, `scopes` etc; I had to name it in the mentioned way because if I named it `description` it will override (or take) the description at the general API info level.

pmorelli92: I can of course change it to `x-description` as well as an alternative :)

ubogdan: Then @securitydefinitions.description should be more appropiate?

pmorelli92: I can do that :)

ubogdan: You made it @security.definition.description instead of @description ...

pmorelli92: Sorry, did not see your edited comment 😅 
I will try to make it work with just `description` but as I said before, when I tried it overlapped with the `@description` that you put to the general API info. I will debug and see if I can solve that.

pmorelli92: The current way it works is the following, given these lines:

```
// @description this API is a demo
// @securitydefinitions.oauth2.application  OAuth2Application
// @description                             Perform OAUTH2 at application level
```

The code will:

- Range every comment
- For the first line it will set the API description to `this API is a demo`
- For the second line it will range all comments starting from `comments[i+1:]` in other words skipping the current. This will effectively go through the `@description` and set it correctly.
- For the third line it will set the API description (and override) to `Perform OAUTH2 at application level`

The issue is that the `parseSecAttr` function is iterating all the lines below until it finds another `@securitydefinitions.`. But then, the lines that `parseSecAttr` iterated, will be iterated back again by the main for buckle.

This is why I changed the name to something different. I can potentially see how to make the main buckle not go through the lines that the security definition already did, but that might introduce lots of changes.

pmorelli92: @ubogdan check it now, with the changes I have done it is working correctly and the main buckle does not re-iterate the lines that the security definitions already iterated :)

ubogdan (dismissed):
LGTM

pmorelli92:
@ubogdan I discovered a bug when you could have more than one security definition so I did the fix and added test cases for it.

ubogdan (approved):
LGTM

ubogdan:
@pmorelli92 Thanks for your contribution.

PR #1181, opened 2022-04-18
Default required

AndreyAndreevich:
**Describe the PR**
Added new option (defaultRequired) for swag init

**Additional context**
I ran into a problem that I have to put validate required next to each field. This is not very convenient when a lot of code has already been written and you have to add the same thing to the tags for each field. I decided that this option will help not only me, but also the community.
Also added the ability to add validate optional fields.
Please merge it)

### DIFF
--- a/cmd/swag/main.go
+++ b/cmd/swag/main.go
@@ -26,6 +26,7 @@ const (
 	codeExampleFilesFlag = "codeExampleFiles"
 	parseInternalFlag    = "parseInternal"
 	generatedTimeFlag    = "generatedTime"
+	defaultRequiredFlag  = "defaultRequired"
 	parseDepthFlag       = "parseDepth"
 	instanceNameFlag     = "instanceName"
 	overridesFileFlag    = "overridesFile"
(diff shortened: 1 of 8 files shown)


mikhaylovv (dismissed):

cmd/swag/main.go
```diff
@@ -26,6 +26,7 @@ const (
 	codeExampleFilesFlag = "codeExampleFiles"
 	parseInternalFlag    = "parseInternal"
 	generatedTimeFlag    = "generatedTime"
+	defaultRequiredFlag  = "defaultRequired"
```
ubogdan, line 29: I think "requiredByDefault" sounds better.

ubogdan (approved):
LGTM

ubogdan:
@AndreyAndreevich I don't like the idea to have a switch that will turn all the generated documentation upside down, but it may be useful to lazy developers. 
Please open an issue describing this feature so we can have evidence of this implementation.

ubogdan:
@AndreyAndreevich we can't merge this due to conflicts.

cmd/swag/main.go
AndreyAndreevich: Fixed

ubogdan (approved):
LGTM

Issue #1171, opened 2022-04-01
(go 1.18) :cannot find type definition: error


suduaya:
**Describe the bug**
Swag generate fails when parsing an `error` type

**To Reproduce**
```
import (
	...
	// Swagger Dependency
	_ "mypackage"
)

 
// MyHandler My handler handles things
// @Success 200 {object} mypackage.SucessResponse
// @Success 500 {object} error
// @Router /v1/myhandler [post]
func (h *Handler) MyHandler(w http.ResponseWriter, r *http.Request) error {
```

`:cannot find type definition: error`
but `mypackage.SucessResponse` is parsed successfully

if I change `error` to `mypackage.ErrorResponse` (which uses `error`), and import it:
```
import (
	...
	// Swagger Dependency
	_ "mypackage"
)

 
// MyHandler My hander handles things
// @Success 200 {object} mypackage.SucessResponse
// @Success 500 {object} mypackage.ErrorResponse
// @Router /v1/myhandler [post]
func (h *Handler) MyHandler(w http.ResponseWriter, r *http.Request) error {
```

`:cannot find type definition: mypackage.ErrorResponse`
but `mypackage.SucessResponse` is parsed successfully

Any clue why this is happening ? :thinking: 

**Expected behavior**
Expecting `error` to be parsed as a response object

**Your swag version**
v1.8.1

**Your go version**
1.18

**Swagger cmd**
`swag init --parseDependency -g /path/to/api.go`

ubogdan:
Can you please share the "mypackage.ErrorResponse" type definition with us.

suduaya:
@ubogdan

```
type ErrorResponse struct {
	error
	message           string
}
```

suduaya:
> Can you please share the "mypackage.ErrorResponse" type definition with us.

@ubogdan

```
type ErrorResponse struct {
	error
	message           string
}
```

Niennienzz:
This might not be the workaround you're looking for.
But if you don't need to marshal the field of the `error` type, the generation works properly.

```go
type ErrorResponse struct {
	Message string `json:"message"`
	Cause   error  `json:"-"`
}
```

suduaya:
@ubogdan 

Totally forgot to export Error type...
Thanks! Solved! :smile:

PR #1188, opened 2022-04-27
Add `error` as a Primitive Type

Niennienzz:
**Describe the PR**

- When the handler is returning a custom error struct type, and it contains the built-in `error` interface like below, the error `:cannot find type definition: error` would occur.

```go
type ErrorResponse struct {
	Message string `json:"message"`
	Cause   error  `json:"cause"`
}
```

- I am thinking, since the `error` interface is also a built-in type, ~and when it marshals to JSON using the interface (instead of any underneath implementation), it should be translated to a string~.
- Please let me know if/where I should add more tests!

-----

**Relation Issue**

https://github.com/swaggo/swag/issues/1171

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -81,7 +81,7 @@ func TransToValidSchemeType(typeName string) string {
 		return NUMBER
 	case "bool":
 		return BOOLEAN
-	case "string":
+	case "string", "error":
 		return STRING
 	}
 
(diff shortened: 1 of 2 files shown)


ubogdan:
@Niennienzz Thanks for your contribution, but we won't make this kind of change because the error type is an interface and doesn't get marshaled by default.   
We may add the feature to skip this field with a warning so "swag" will not end with an error, but this is entirely another story.

ubogdan:
Please let me know if you are interested on making the change described by me or we can close this PR.

Niennienzz:
> Please let me know if you are interested on making the change described by me or we can close this PR.

Please close for now. I would love to, but currently I am a bit busy with other projects...

Issue #1131, opened 2022-02-09
Global type override in .swaggo file not working


bmpotter:
**Describe the bug**
First, thank you for this excellent swaggo project! It is a great productivity enhancement.

The `replace` statement in `.swaggo` does not seem to be honored for an external type. I can successfully use `swaggertype` for every use of that external type in all of my structs, but i want to be able to define the alternate type once in the `.swaggo` file instead.

**To Reproduce**
Use this `main.go` file:

```
package main

import (
	"fmt"
	"github.com/lib/pq"
)

type MyStruct struct {
	Foo    string         `json:"foo"`    // some field
	Labels pq.StringArray `json:"labels"` // external type not found in .swaggo file

	//Labels pq.StringArray `json:"labels" swaggertype:"array,string"` // this works
}

// @Router /myroute [get]
// @Summary Get something
// @id get
// @Tags MyStruct
// @Produce json
// @Success 200 {object} MyStruct "Resource returned"
func Get() {
	fmt.Println("getting something")
}

// @title My test case
// @version 0.1
// @tag.name MyStruct
func main() {
	fmt.Println("hello")
}
```

And use this `.swaggo` file:

```
// swag does not pick this up, still get msg: cannot find type definition: pq.StringArray
replace github.com/lib/pq.StringArray []string

// Also tried these:
// replace lib/pq.StringArray []string
// replace pq.StringArray []string
// skip github.com/lib/pq.StringArray
```

Then run `$GOPATH/bin/swag init` and you'll get this output:

```
2022/02/09 06:32:08 Using overrides from .swaggo
2022/02/09 06:32:08 Generate swagger docs....
2022/02/09 06:32:08 Generate general API Info, search dir:./
2022/02/09 06:32:08 Generating main.MyStruct
2022/02/09 06:32:08 ParseComment error in file /Users/bp/src/POCs/swaggo-file-bug/main.go :cannot find type definition: pq.StringArray
```

**Expected behavior**
It should run without error, and in the generated `docs/swagger.*` files it should define the `labels` field as a string array.

**Your swag version**
The same problem occurs in both:
v1.7.8
v1.7.9-p1

**Your go version**
1.17.5

**Desktop:**
MacOS Big Sur 11.6

ubogdan:
Hi, this is a quite new feature and for sure it's not perfect. This is also an open-source project and everything was realized with the community contribution. 
If you find this feature useful and you have some spare time feel free to contribute and we will be more than happy to assist you with a code review.

[...]

PR #1212, opened 2022-05-22
Implement error (and any) like interface

sapk:
This manage error and any like interface. 

Previously, any would result in a primitive type `any` which seem not valid in swagger definition.

This basically, set an empty definition like for `interface{}` previously for `any` and `error` types.

**Relation issue**
- https://github.com/swaggo/swag/pull/1188
- https://github.com/swaggo/swag/issues/1171
- https://github.com/swaggo/swag/issues/1131

**Additional context**
This fix cases when you cannot override with tag `swaggertype` in case the struct is from an external dependency. 
Previously this case need to redefine the struct locally to override the field type but can be cumbersome in case of deeply nested struct and It is also not possible to override go types with  .swaggo file.

I first try to implement like suggested in [this comment](https://github.com/swaggo/swag/pull/1188#issuecomment-1111008162) but after reviewing some cases (like in the testdata test) this feel better and more logical to consider error (and any) like interface. Let me know if it doesn't fit you.

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -868,6 +868,9 @@ func convertFromSpecificToPrimitive(typeName string) (string, error) {
 }
 
 func (parser *Parser) getTypeSchema(typeName string, file *ast.File, ref bool) (*spec.Schema, error) {
+	if IsInterfaceLike(typeName) {
+		return &spec.Schema{}, nil
+	}
 	if IsGolangPrimitiveType(typeName) {
 		return PrimitiveSchema(TransToValidSchemeType(typeName)), nil
 	}
--- a/gen/gen_test.go
+++ b/gen/gen_test.go
@@ -779,3 +779,45 @@ func TestGen_Debugger(t *testing.T) {
 		_ = os.Remove(expectedFile)
 	}
 }
+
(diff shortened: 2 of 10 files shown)


parser.go
```diff
@@ -868,6 +868,9 @@ func convertFromSpecificToPrimitive(typeName string) (string, error) {
 }
 
 func (parser *Parser) getTypeSchema(typeName string, file *ast.File, ref bool) (*spec.Schema, error) {
+	if IsInterfaceLike(typeName) {
+		return &spec.Schema{}, nil
+	}
```
ubogdan, line 873: We need to cover this with unit tests.

ubogdan (changes requested):
Good job! Please write unit test to cover new added functionality

parser.go
sapk: It should already be covered by `TestGen_ErrorAndInterface` and `TestIsInterfaceLike` but I can recheck.

sapk: I added a test similar to TestGen_ErrorAndInterface base on testdata in the root package.

ubogdan (approved):
LGTM

Issue #1213, opened 2022-05-23
go generics cannot parse


medivh-jay:
**Describe the bug**
An error occurred while parsing a data structure that contains generics.

```
ParseComment error in file news.go :unknown field type &ast.IndexListExpr{X:(*ast.SelectorExpr)(0xc00047a750), Lbrack:267, Indices:[]ast.Expr{(*ast.StarExpr)(0xc00047a780), (*ast.SelectorExpr)(0xc00047a798)}, Rbrack:301}
```

This is my code:
```
type Filter interface {
	~func(selector *sql.Selector)
}

type query[T any, F Filter] interface {
	Where(ps ...F) T
}

type Pager[T query[T, F], F Filter] struct {
	Rows   uint8                `json:"rows" form:"rows"`
	Page   int                  `json:"page" form:"page"`
	NextID *valuer.StringValuer `json:"next_id" form:"next_id"`
	PrevID *valuer.StringValuer `json:"prev_id" form:"prev_id"`
	query  T
}

type NewsPager struct {
	request.Pager[*entity.NewsQuery, predicate.News]
	Search string `json:"search" form:"search"`
}
```

and this is my comment:
```
// @Summary xxx
// @Tags xxx
// @Description xxx
// @Param   data query  request.NewsPager true ""
// @Success 200 {object} response.Response{data=newsResponse} "ok"
// @Router  /api/v1/news [get]
```

**Your swag version**
e.g. 1.8.2

**Your go version**
1.18

**Desktop (please complete the following information):**
 - OS: Windows 11

medivh-jay:
create a project directly using the following code:
```
package main

type NewsSelector func(selector func())

type Filter interface {
	~func(selector func())
}

type query[T any, F Filter] interface {
	Where(ps ...F) T
}

type Pager[T query[T, F], F Filter] struct {
	Rows   uint8   `json:"rows" form:"rows"`
	Page   int     `json:"page" form:"page"`
	NextID *string `json:"next_id" form:"next_id"`
	PrevID *string `json:"prev_id" form:"prev_id"`
	query  T
}

type String string

func (String) Where(ps ...NewsSelector) String {
	return ""
}

type NewsPager struct {
	Pager[String, NewsSelector]
	Search string `json:"search" form:"search"`
}

// @Summary xxx
// @Tags xxx
// @Description xxx
// @Param   data query  NewsPager true "1"
// @Success 200 {object} string "ok"
// @Router  /api/v1/news [get]
func route() {

}

[...]

PR #1214, opened 2022-05-25
Suppress generic parsing errors

medivh-jay:
**Describe the PR**
Suppress generic parsing errors

**Relation issue**
https://github.com/swaggo/swag/issues/1213


medivh-jay:
The lower version of go cannot be compiled

PR #1219, opened 2022-06-02
feat: parse only specific extension tag

igorkova:
**Describe the PR**
Enable parsing (output) of specific operations which have specific tag (extension).

**Relation issue**
https://github.com/swaggo/swag/issues/661

**Additional context**


igorkova:
This is not 100% ready yet. Let us get back

sid-rpf:
This is a nice and must-have feature. 👍

ubogdan:
@igorkova let me know when you are done with this.

igorkova:
@ubogdan This is done as far as we are concerned.
It covers all of our needs

Example usage will be `swag --parseExtension google-backend` which will take all routes which have `// @x-google-backend` comment and export them in a separate file.
That file we can then take and forward to ESPV2 and by doing that we can control which endpoints are exposed on the gateway (or any other external openAPI definition) vs those which will be exposed internally

igorkova:
@ubogdan great if we can get workflow approved  🙌

parser.go
```diff
@@ -828,12 +838,29 @@ func (parser *Parser) matchTags(comments []*ast.Comment) (match bool) {
...
+
+			if lowerAttribute == fmt.Sprintf("@x-%s", strings.ToLower(parser.parseExtension)) {
+				return true
+			}
+		}
+		return false
+	}
+	return true
+}
+
```
ubogdan, line 856: The only thing you are using from the parser is parser.ParseExtension, This function is a mimic of in_array `func in_array(val string, array []string) bool` with default to true if the search value is empty. 

Such a simple function should not be attached to the parser.
```go
func matchExtension(textToFind string, comments []*ast.Comment) bool
```

ubogdan (changes requested):

parser.go
igorkova: Got it.
I was basically c/p above matchTags without too much thinking. you have a good point.

I can detach it from the parser, before that just to double check do you recommend to create a "helper" in_array-type method or still keep it specific to extension parsing? (Not sure if we have such a helper file/methods today somewhere - can't find them in the repo)

ubogdan: Just detach the function from the parser.

ubogdan: Thanks.

ubogdan (approved):

ubogdan:
@igorkova Thanks for your contribution.

PR #1238, opened 2022-06-20
fix: don't skip ".." directory

haoqixu:
`Parser.Skip()` treats `..` directory as hidden folder and return a `filepath.SkipDir`. If `..` is provided via `--dir` option, `swag` will miss some files.


sdghchj (approved):
LGTM

Issue #1179, opened 2022-04-18
form is not supported paramType


libear:
<img width="1117" alt="image" src="https://user-images.githubusercontent.com/2178420/163788824-2a58130a-7bc9-4496-94e0-447406c2f819.png">

ubogdan:
And it supposed to be?

ubogdan:
OAS2 supports form encoded payload via "multipart/form-data" and this is implemented in swagger via
`// @accept mpfd` see [mime-types](https://github.com/swaggo/swag#mime-types) and 
`// @Param   example     formData     string     false  "description"  ` see [param-type](https://github.com/swaggo/swag#param-type)

The same is referenced here in [OAS2 specs](https://swagger.io/docs/specification/2-0/file-upload/)..

ismdeep:
@ubogdan swagger may meet some base modules which has `form` but we do not support. In this case, the swag program will crash with this message.

```text
2022/08/16 09:51:59 ParseComment error in file /opt/go/pkg/mod/github.com/astaxie/beego@v1.12.1/parser.go :form is not supported paramType
make: *** [swag-doc] Error 1
```

`/opt/go` is `GOPATH`

PR #1243, opened 2022-06-25
fix: form is not supported paramType

ismdeep:
**Relation issue**
https://github.com/swaggo/swag/issues/1179

### DIFF
--- a/operation.go
+++ b/operation.go
@@ -392,6 +392,12 @@ func (operation *Operation) ParseParamComment(commentLine string, astFile *ast.F
 
 			param.Schema = schema
 		}
+	case "form":
+		if objectType == PRIMITIVE {
+			param.Schema = PrimitiveSchema(refType)
+		} else {
+			param.Schema = nil
+		}
 	default:
 		return fmt.Errorf("%s is not supported paramType", paramType)
 	}


operation.go
```diff
@@ -392,6 +392,12 @@ func (operation *Operation) ParseParamComment(commentLine string, astFile *ast.F
 
 			param.Schema = schema
 		}
+	case "form":
```
ubogdan, line 395: We have formData param type. Why do we need this ?

ismdeep: The problem was caused by github.com/astaxie/beego v1.12.1

https://github.com/astaxie/beego/blob/de5650b7230083f957045890af942ffdac87c782/parser.go :377

Which I suppose we would be compatible to this situation.

ismdeep: > We have formData param type. Why do we need this ?

The problem was caused by github.com/astaxie/beego v1.12.1

https://github.com/astaxie/beego/blob/de5650b7230083f957045890af942ffdac87c782/parser.go :377

Which I suppose swag would be compatible to this situation.

ismdeep: @ubogdan hi, the problem was caused by swag meets github.com/astaxie/beego.

```text
ParseComment error in file /src/server/vendor/github.com/astaxie/beego/parser.go :form is not supported paramType
```

ubogdan: We need a code example to understand the issue better.
After we agree with the changes, you will have to provide unit test for new change/functionality.

ismdeep: This problem is caused by a upstream package, the swag program meet it will crashed. I suggest add this `form` type or skip the crash with an empty value.

:( `swag` program crashed that we can NOT update docs folder in my project.

Thanks.

ubogdan:
Don't get me wrong. You can't come up with a solution to an issue that you can explain and expect us to belive you. 

1. Please come up with an example where we can run swag and generate a crash. 
2. If you intend to contribute with a fix, we need to agree on the implementation and write unit tests to cover new code changes.

ismdeep:
> Don't get me wrong. You can't come up with a solution to an issue that you can't explain and expect us to believe you.
> 
> 1. Please come up with an example where we can run swag and generate a crash.
> 2. If you intend to contribute with a fix, we must agree on the implementation and write unit tests to cover new code changes.

Here is an example to recur the swag crash problem. [https://github.com/ismdeep/swag-crash-demo-001](https://github.com/ismdeep/swag-crash-demo-001)

For more information about how to recur the problem, please see the [README.md](https://github.com/ismdeep/swag-crash-demo-001/blob/main/README.md).

Issue #1150, opened 2022-02-25
Wrong formating for custom struct names


F0rzend:
**Wrong formating for custom struct names**
When run `swag fmt` 2 spaces appear between `@name` and the name instead of one and the name is treated as an empty string

**To Reproduce**
1. Write a struct, for example
```go
type User struct {
	ID       uint   `json:"id"`
	Username string `json:"username"`
} // @name User
```
2. Run `swag fmt -g filename.go`
3. Get `@name  User` instead of `@name User` in your code
4. Run `swag init -g path/to/mani.go`
5. In docs/swagger.yaml you will see
```yaml
definitions:
  "":
    description: User information
    properties:
      id:
        type: integer
      username:
        type: string
    type: object
```

**Expected behavior**
Correct type name, "User"

**Screenshots**
![image](https://user-images.githubusercontent.com/50886859/155784641-86b0c7dc-2d74-4372-a70c-ff311ca5ed7e.png)

**Swag version**
```sh
$ swag -v
swag version v1.8.0
```

**Go version**
```sh
$ go version
go version go1.18beta1 linux/amd64
```

**Desktop:**
 - OS: OS: Ubuntu 21.10 x86_64 
 - Browser: Firefox browser 97.0 (64-bit); Mozila Firefox for Ubuntu canonical - 1.0
 - Version: 21.10

ubogdan:
This is an open-source project. Feel free to contribute with a fix and I will be more than happy to assist you with a Code Review.

PR #1248, opened 2022-06-30
fix: fmt issue with struct

MattKetmo:
Fix issue https://github.com/swaggo/swag/issues/1150

To summarize the issue, running `swag fmt` will set (at least) 2 spaces between the attribute tag and the rest of the comment. But this is a problem for struct with single comment:

```diff
type Example struct {
	Foobar string
-} // @name Example
+} // @name  Example
```

Indeed the doc generation doesn't interpret this correctly for struct (it completely ignore it).

So there are 2 solutions:

- fix the indent (which is feel more natural)
- fix the doc generation to correctly parse 2 spaces

This commit is fixing the indent part.

Basically I just set the padding=1 on `tabwriter`.

The cons is that it will also impacts all other godocs.

For instance (taken from the readme):

```diff
-// @title           Swagger Example API
-// @version         1.0
-// @description     This is a sample server celler server.
-// @termsOfService  http://swagger.io/terms/
+// @title          Swagger Example API
+// @version        1.0
+// @description    This is a sample server celler server.
+// @termsOfService http://swagger.io/terms/
```

I don't know if there was a reason for the choice of padding=2 apart aesthetic.
Ping @HYY-yu who first use it in commit [93acd2ad](https://github.com/swaggo/swag/commit/93acd2ad51e848ef3c78b25cb3964745fecde054)

### DIFF
--- a/formatter.go
+++ b/formatter.go
@@ -78,7 +78,7 @@ func formatComments(fileName string, contents []byte, formattedComments []byte, 
 }
 
 func formatFuncDoc(commentList []*ast.Comment, formattedComments io.Writer, oldCommentsMap map[string]string) {
-	w := tabwriter.NewWriter(formattedComments, 0, 0, 2, ' ', 0)
+	w := tabwriter.NewWriter(formattedComments, 0, 0, 1, ' ', 0)
 
 	for _, comment := range commentList {
 		text := comment.Text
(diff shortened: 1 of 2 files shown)


ubogdan (approved):
LGTM

ubogdan:
@MattKetmo Thanks for your contribution.

PR #1252, opened 2022-07-06
1) add logrus as logger package, so we can handle more log level

extrame:
**Describe the PR**

1. add logrus as logger package, so we can handle more log level
2. change parser.debuger to parser.logger, so use it to log other levels make sense.
3. make more log use logger.trace. 
4. add verbose flag, so user can see more detail of executation( trace level).
5. allow search unnamed import even if parseDependency is disabled. (reason: unnamed import is used widely; not all dependency need parsed, just need those which is used in comments).

**Relation issue**
no

**Additional context**
no


MrDOS:
As a consumer of this library, I'm generally leery of libraries tying themselves to any particular logging library. While most of my applications do depend on Logrus, I'd hate for an application which uses something else – e.g., the standard logger, [zap](https://github.com/uber-go/zap), or [Zerolog](https://github.com/rs/zerolog) – to be saddled with a transitive dependency on Logrus.

At the same time, I understand the desire to avail oneself of more structured logging in the `swag` CLI application. Perhaps the swag CLI could be moved into its own module so that it can depend on Logrus independently of the top-level library?

PR #1261, opened 2022-07-12
Test host && make combined object definition can based on interface{} && add one more tips

extrame:
**Describe the PR**
1) test host definition
2) give more information if param type is wrong
3) make combined object definition can defined based on interface{} like:
" @Success		200    {object}  {success=bool}"

**Relation issue**
none

**Additional context**
none


ubogdan:
@extrame We expect all the contributions to be based on an open issue and to discuss the directions, so the developer's work wont is in vain. 
 I want to remind you that OpenApi Documentation (swagger) is about documenting types, not values. 

You are documenting the response value for HTTP 200 OK if I understand well.

Issue #1240, opened 2022-06-24
Get Swagger instance from register


rytsh:
If documentation generated by swaggo and if we added that generated docs.go file to the git ignore list, we cannot edit SwaggerInfo without run swaggo.

I want to edit SwaggerInfo(version, appname..), if "swagger" instance exists.

There is already `Register` method exists and can we also add `GetInstance` and get back Swagger from there?

PR #1298, opened 2022-08-17
feat: get swagger instance from register

rytsh:
**Describe the PR**
Problem starts if our swagger instance generated by CI and registered.
In main program if we want to modify some parameters inside of that instance is it not possible because code not in there.

Getting swagger instance with name from register it is easy to work with it, without need any workaround.

Swagger is an interface so we should do some casting but we know the our code and it should be this type.

```go
if s, ok := swag.GetSwagger("swagger").(*swag.Spec); s != nil && ok {
	s.Version = "1.0.0"
	s.Title = "Test"
}
```

**Relation issue**
#1240

### DIFF
--- a/swagger.go
+++ b/swagger.go
@@ -39,6 +39,15 @@ func Register(name string, swagger Swagger) {
 	swags[name] = swagger
 }
 
+// GetSwagger returns the swagger instance for given name.
+// If not found, returns nil.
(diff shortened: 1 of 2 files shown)


ubogdan (approved):
LGTM

ubogdan:
@rytsh Thanks for your contribution.

PR #1313, opened 2022-08-30
refactor func splitStructName

sdghchj:
Just reimplement the func splitStructName with std functions in library `strings`

### DIFF
--- a/generics.go
+++ b/generics.go
@@ -10,6 +10,7 @@ import (
 	"go/ast"
 	"strings"
 	"sync"
+	"unicode"
 )
 
 var genericDefinitionsMutex = &sync.RWMutex{}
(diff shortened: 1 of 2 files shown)


generics.go
```diff
@@ -177,39 +178,40 @@ func (pkgDefs *PackagesDefinitions) parametrizeStruct(file *ast.File, original *
...
-
-		insideBrackets += numOpened - numClosed
-		lastParam += p + ","
-
-		if insideBrackets == 0 {
-			genericParams = append(genericParams, strings.TrimSpace(strings.TrimRight(lastParam, ",")))
-			lastParam = ""
+	depth := 0
+	genericParams = strings.FieldsFunc(genericParams[1], func(r rune) bool {
+		if r == '[' {
```
ubogdan, line 204: It's more elegant with a switch case 

```go
	switch r {
	case '[':
		depth++
	case ']':
		depth--
	case ',':
		if depth == 0 {
			return true
		}
	default:
		return false
	}
```

ubogdan (approved):
LGTM

Issue #866, opened 2021-01-08
upload multi file  is wrong


hanyunmuyu:
upload multi file  is wrong

```

// @Summary 多文件上传
// @Security ApiKeyAuth
// @Description | 参数 | 说明 |备注|
// @Description | :-----: | :----: | :----: |
// @accept x-www-form-urlencoded
// @Tags  admin
// @version 1.0
// @Param file formData []file true "文件" collectionFormat(multi)
// @success 200 object utils.JSONResult{} "成功"
// @Router /admin/v1/upload/multi [POST]

```

```
2021/01/08 11:34:52 ParseComment error in file app/http/controllers/admins/upload_controller.go :file is not supported array type for formData


```

ubogdan:
If you made a variable `var files []file` somewhere this file does your code still compile :) .

fr3fou:
Is there a solution to this yet?

MuhammedOzdogan:
What's the current status of this feature?
#1332 is merged but `[]file` does nothing different than `file` or I couldn't manage to make it work in my local machine.
Any change to update documentation, I would appreciate if anyone adds an example under `/example` dir

SeanTolstoyevski:
We are also wondering what the status of this issue is.

We have APIs for users to upload multiple files, but only a single file can be selected from the swagger web interface.

If there is no solution for this issue, you can at least explain what the problem is and encourage a developer to solve it.

example swagger docs:

```go
// .......................
// @Tags Attachment
// @Accept  multipart/form-data
// @Produce  json
// @Security BearerAuth
// @Param files formData file true "S3'e yuklenecek olan dosyalar"
// @Param is_private query string true "..."
// @Param folder query string true "..."
// @Success 200 {array} models.AWSContent "...."
```

InfernoZero:
Try below:
It worked well for me :)

// @Param files formData []file true "....." collectionFormat(multi)

kingindanord:
1+

PR #1332, opened 2022-09-18
feat. add support formData []file

otakukaze:
**Describe the PR**
add support formData multiple file

**Relation issue**
https://github.com/swaggo/swag/issues/866

### DIFF
--- a/operation.go
+++ b/operation.go
@@ -226,7 +226,7 @@ func findInSlice(arr []string, target string) bool {
 }
 
 func (operation *Operation) parseArrayParam(param *spec.Parameter, paramType, refType, objectType string) error {
-	if !IsPrimitiveType(refType) {
+	if !IsPrimitiveType(refType) && !(refType == "file" && paramType == "formData") {
 		return fmt.Errorf("%s is not supported array type for %s", refType, paramType)
 	}
 


operation.go
```diff
@@ -226,7 +226,7 @@ func findInSlice(arr []string, target string) bool {
 }
 
 func (operation *Operation) parseArrayParam(param *spec.Parameter, paramType, refType, objectType string) error {
-	if !IsPrimitiveType(refType) {
+	if !IsPrimitiveType(refType) && !(refType == "file" && paramType == "formData") {
```
ubogdan, line 229: You are cheating the validation because `file` is not a primitive and I won't allow this hack.

What definition will generate this change if I define type `file` as the following?
```go
type file []string
```

otakukaze: Screen Shot 2022-09-19 at 11 50 33 AM

ubogdan: I agree this should work as it was provided in the example above but only for`@consume x-www-form-urlencoded` or `@consume application/x-www-form-urlencoded`. 

A mix of `in:formData` with `in:Body` params should not be allowed.

ubogdan (approved):
LGTM

Issue #1353, opened 2022-10-20
Swaggo ignores package alias


doublebigguy:
Swaggo ignores package alias, as I can see, it tries to generate from struct that in create package

// @Success     201     object views.JsonApiView[views.JsonApiNode[createView.JsonApiAttributes]]

2022/10/20 07:40:57 Generating $views.JsonApiView-views_JsonApiNode_create_JsonApiAttributes
2022/10/20 07:40:57 Generating $views.JsonApiNode-create_JsonApiAttributes

sdghchj:
Try the latest master branch, please.

doublebigguy:
It looks good, thanks
import (
_ "ordering/internal/gateways/http/views/api/sets/v1/create" // Here is JsonApiAttributesV1
_ "ordering/pkg/views" // Here is generic
)
and in another the same but products instead of sets
import (
_ "ordering/internal/gateways/http/views/api/products/v1/create" // Here is JsonApiAttributesV1
_ "ordering/pkg/views" // Here is still generic
)
The result is
#/definitions/views.JsonApiView-views_JsonApiNode-ordering_internal_gateways_http_views_api_products_v1_create_JsonApiAttributesV1-any
#/definitions/views.JsonApiView-views_JsonApiNode-ordering_internal_gateways_http_views_api_sets_v1_create_JsonApiAttributesV1-any

doublebigguy:
UPD:
But if i have, smth like this in imports

"ordering/internal/gateways/http/presenters/api/products/v1/create"
_ "ordering/internal/gateways/http/views/api/products/v1/create" // here is JsonApiAttributesV1

It failes with message ParseComment error in file ordering\application\internal\gateways\http\controllers\api\products\v1\create.go :cannot find type definition: create.JsonApiAttributesV1

doublebigguy:
UPD: resolved by using _.JsonApiAttributesV1 instead of create.JsonApiAttributesV1

sdghchj:
> UPD: But if i have smth like this in imports
> 
> "ordering/internal/gateways/http/presenters/api/products/v1/create" _ "ordering/internal/gateways/http/views/api/products/v1/create" // here is JsonApiAttributesV1
> 
> It failes with message ParseComment error in file ordering\application\internal\gateways\http\controllers\api\products\v1\create.go :cannot find type definition: create.JsonApiAttributesV1

swag will search for package with the same name from the imports sequencely, so in this case, first imported package will be selected

[...]

PR #1360, opened 2022-10-25
Fix generics with package alias

sdghchj:
**Describe the PR**
fix generics bug:
1. More than one generics types actually come from different packages, but with the same shot package name and type name.
2. Import generics types from a package with an alias.

**Relation issue**
#1353


ubogdan (approved):
LGTM

Issue #1367, opened 2022-10-27
I would like to define enum.


KousukeUchiyama:
What is the best way to output the following Swagger?

```yaml
swagger: "2.0"

components:
  schemas:
    EnumFileStatus:
      type: string
      title: EnumFileStatus
      enum:
        - not_uploaded
        - file_error
        - data_error
        - uploaded
    ResponseUploadFile:
      title: ResponseUploadFile
      type: object
      description: ''
      properties:
        fileStatus:
          $ref: '#/components/schemas/EnumFileStatus'
```

I would like to define enum.
We have tried the following methods, but did not achieve the intended result.

https://github.com/swaggo/swag#attribute

sdghchj:
global enums haven't been supported yet

Zxilly:
Any plan supporting this?

sdghchj:
I would like give an implementation.

for example:
```
package types

type Class int

const (
	A Class = iota + 1 // AAA
	B                  /* BBB */
	C
	D
)

type Mask int

const (
	Mask1 Mask = 1 << iota // Mask1
	Mask2                  /* Mask2 */
	Mask3                  // Mask3
	Mask4                  // Mask4
)

type Type string

const (
	Teacher Type = "teacher" // teacher
	Student Type = "student" /* student */
	Other   Type = "Other"   // Other
	Unknown      = "Unknown"
	OtherUnknown
)

type Person struct {
	Name  string
	Class Class
	Mask  Mask
	Type  Type
}
```

this will output:
```
definitions:
  types.Class:
    type: integer
    enum:
      - 1
      - 2
      - 3
      - 4
    x-enum-comments:
      A: AAA
      B: BBB
    x-enum-varnames:
      - A
      - B
      - C
      - D
  types.Mask:
    type: integer
    enum:
      - 1
      - 2
      - 4
      - 8
    x-enum-comments:
      Mask1: Mask1
      Mask2: Mask2
      Mask3: Mask3
      Mask4: Mask4
    x-enum-varnames:
      - Mask1
      - Mask2
      - Mask3
      - Mask4
  types.Person:
    type: object
    properties:
      class:
        $ref: '#/definitions/types.Class'
      mask:
        $ref: '#/definitions/types.Mask'
      name:
        type: string
      type:
        $ref: '#/definitions/types.Type'
  types.Type:
    type: string
    enum:
      - teacher
      - student
      - Other
    x-enum-comments:
      Other: Other
      Student: student
      Teacher: teacher
[...]

PR #1387, opened 2022-11-19
parse global enums

sdghchj:
**Describe the PR**
if a global custom type is defined based on an underlying integer or string type, it will be considered as an enum type.

an enum type exists with two extensions:
x-enum-varnames:  record their names
x-enum-comments:  record their comments associated to their names.

**Relation issue**
#1367


sdghchj:
@ubogdan

const.go
```diff
@@ -0,0 +1,122 @@
...
+		if cv.Type == nil && evalType != nil {
+			cv.Type = evalType
+		}
+		if value != nil {
+			cv.Value = value
+		}
+		return value
+	}
+	return cv.Value
+}
```
ubogdan, line 30: The behavior of this function is quite unusual. it value is an ast.Expresion , it populates the ConstVariable type and value , otherwise, it returns a value that may be ni ?

ubogdan (approved):
The implementation was written in a hurry, I

const.go
sdghchj: cv.Value is initialized in the func `collectConstVariables` with a ast.Expr, and then will be reassigned with the evaluated non-nil value of integer or string type，so it would never be nil.

PR #1396, opened 2022-11-24
support json tags in embedded structs

### LINKED ISSUE #886: Support embedded struct with json tag
Currently If you have an embded struct with a json tag, the json tag is ignored by swaggo. When marshaling to json normally it is not.
```
package main

import "encoding/json"
import "fmt"

type A struct {
    Foo string
}

type B struct {
    A `json:"A"`
    Bar string
}

func main() {
    b, _ := json.Marshal(B{})
    fmt.Println(string(b))
}
```

output (should also be the expected result for swaggo):
```
{"A":{"Foo":""},"Bar":""}
```

Currently with swaggo generation the result would be
```
{"Foo":"", "Bar":""}
```
which is not expected

When using a custom type that has a embeded struct it should still read the json tag and generate the definition the same as my expected result above.

jgillich:
**Describe the PR**
This change makes it so json tags on embedded structs are honored.

One example of this:
```go

type RevValueBase struct {
	Status bool `json:"Status"`

	Err int32 `json:"Err,omitempty"`
}
type RevValue struct {
	RevValueBase `json:"rev_value_base"`
}
```

Previously, `RevValueBase` was inlined into `RevValue`. With this change, it matches the behavior of the core json package and the embedded struct becomes a field.
```json
"rev_value_base": {
    "$ref": "#/definitions/web.RevValueBase"
}
```

It should be noted that this changes the schema generation of embedded structs. However, since it was impossible to unmarshal embedded structs previously, I don't believe this is could cause breakage unless people implemented custom de-serialization.

**Relation issue**
Fixes #886

### DIFF
--- a/field_parser.go
+++ b/field_parser.go
@@ -43,7 +43,7 @@ func newTagBaseFieldParser(p *Parser, field *ast.Field) FieldParser {
 
 func (ps *tagBaseFieldParser) ShouldSkip() bool {
 	// Skip non-exported fields.
-	if !ast.IsExported(ps.field.Names[0].Name) {
+	if ps.field.Names != nil && !ast.IsExported(ps.field.Names[0].Name) {
 		return true
 	}
 
@@ -74,6 +74,10 @@ func (ps *tagBaseFieldParser) FieldName() (string, error) {
 		if name != "" {
 			return name, nil
 		}
+	}
+
+	if ps.field.Names == nil {
+		return "", nil
 	}
 
 	switch ps.p.PropNamingStrategy {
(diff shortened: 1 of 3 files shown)


sdghchj:
LGTM， But I'd like to follow @ubogdan 's opinion for this feature.

ubogdan (approved):
LGTM

ubogdan:
> LGTM， But I'd like to follow @ubogdan 's opinion for this feature.

As soon as it doesn't break older contract I'm fine.

ubogdan:
@jgillich Thanks for your contribution.

Issue #1401, opened 2022-11-28
Cannot add example tag containing double colon (":")


GusAntoniassi:
**Describe the bug**
The following struct fails to generate the docs through `swag init`:

```golang
type Foo struct {
  Attributes map[string]string `example:"bucket_arn:arn:aws:::foobar-bucket"`
}
```

The following error is returned:
```
ParseComment error in file /tmp/foobar/main.go :example value bucket_arn:arn:aws:::foobar-bucket should format: key:value
```

Escaping the value with double backlashes (`\\`) does not work.

**To Reproduce**
1. Create the following `main.go` file:
```golang
package main

type Resource struct {
        Attributes    map[string]string `example:"bucket_arn:arn:aws:::foobar-bucket"`
}

//      @Summary                Foo
//      @Success                200     {object}        Resource
//      @Router                 /foo [get]
func main() {

}
``` 
2. Generate the docs with `swag init ./main.go`

**Expected behavior**
Users should be able to create examples with double colons, maybe escaping them with double backlashes to avoid key:value conflicts.

**Your swag version**
1.8.8

**Your go version**
e.g. 1.19.3

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: Not relevant

PR #1402, opened 2022-11-28
chore: parse escaped double colon (\\:) example struct tag

GusAntoniassi:
**Describe the PR**
Replaces the regular string splitting by a ":" made during the "example" tag parsing so that it will not split escaped double colons, allowing string examples containing a double colon to be escaped using double backlashes.

**Relation issue**
https://github.com/swaggo/swag/issues/1401

**Additional context**
Since the core `regexp` library cannot suport negative lookaheads, I had to combine a regular expression with a loop to capture all the ocurrences and append them to an array.

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -15,7 +15,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"reflect"
-	"regexp"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


sdghchj:
I see your purpose is to make the map value contain colons.
But I wonder why not use a light mode to fix it, for example, use `strings.IndexRune` to find the first colon and split the string into two parts, a key and a value.

GusAntoniassi:
Hello @sdghchj, initially I went with this approach to allow escaped double colons in the key value, but thinking better about it it does not make much sense and I can't think of a valid use case for it.

I've refactored the code using `strings.SplitN` to split at the first ocurrence of a `:`, this makes the change much simpler and from my tests it won't be necessary to escape the remaining double colons in the value, so for example the struct from my issue would work without any changes:

```golang
type Foo struct {
  Attributes map[string]string `example:"bucket_arn:arn:aws:::foobar-bucket"`
}
```

sdghchj (approved):

PR #1406, opened 2022-12-05
fix: string cutset method usage

ismdeep:
**Describe the PR**
Here we are going to trim `//`, `/*` at the comment before and `*/` last, should use `TrimPrefix` and `TrimSuffix` instead of `TrimLeft` and `TrimRight`.

**Relation issue**
None

### DIFF
--- a/packages.go
+++ b/packages.go
@@ -372,9 +372,9 @@ func (pkgDefs *PackagesDefinitions) collectConstEnums(parsedSchemas map[*TypeSpe
 			}
 			if constVar.Comment != nil && len(constVar.Comment.List) > 0 {
 				enumValue.Comment = constVar.Comment.List[0].Text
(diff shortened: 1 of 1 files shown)


ubogdan:
Please add unit tests that "ensure" this change is genuine and won't be reverted by mistake at a later time.

sdghchj:
> Please add unit tests that "ensure" this change is genuine and won't be reverted by mistake at a later time.

No need actually, it has already been covered by tests.

But, here we have to discuss, if a const with comment like

```
type Type string
const Teacher Type = "teacher" //// teacher
```

whether the invalid comment is `teacher` or `// teacher` ?

Maybe TrimPrefix and TrimSuffix are more accurate than TrimLeft and TrimRight.

The standard lib has provide a method to extract the comment text, but with a `\n` at the end.
```
func (g *CommentGroup) Text() string{
  ...
}
```

sdghchj (approved):

PR #1422, opened 2022-12-14
Use form tags when available

pconstantinou:
**Describe the PR**
swaggo doesn't look at the `form` tag on structs to determine field names, only the `json` tag. This update sets the field name in the swagger files to the `form` tag if it exists. This will match what gin binds to.

**Relation issue**
This issue was also raised here:
https://github.com/swaggo/swag/discussions/1411

**Additional context**

### DIFF
--- a/field_parser.go
+++ b/field_parser.go
@@ -68,16 +68,16 @@ func (ps *tagBaseFieldParser) ShouldSkip() bool {
 func (ps *tagBaseFieldParser) FieldName() (string, error) {
 	var name string
 
-	// use "form" tag over json tag
-	name = strings.TrimSpace(strings.Split(ps.tag.Get(formTag), ",")[0])
-	if name != "" {
-		return name, nil
-	}
-
 	if ps.field.Tag != nil {
 		// json:"tag,hoge"
 		name = strings.TrimSpace(strings.Split(ps.tag.Get(jsonTag), ",")[0])
 
+		if name != "" {
+			return name, nil
+		}
+
+		// use "form" tag over json tag
+		name = strings.TrimSpace(strings.Split(ps.tag.Get(formTag), ",")[0])
 		if name != "" {
 			return name, nil
 		}
--- a/field_parser.go
+++ b/field_parser.go
@@ -67,6 +67,13 @@ func (ps *tagBaseFieldParser) ShouldSkip() bool {
 
 func (ps *tagBaseFieldParser) FieldName() (string, error) {
 	var name string
+
+	// use "form" tag over json tag
+	name = strings.TrimSpace(strings.Split(ps.tag.Get(formTag), ",")[0])
+	if name != "" {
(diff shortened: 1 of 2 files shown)


ubogdan:
Why do we want to introduce such a fallback?

pconstantinou:
The swagger file that's currently generated is incorrect. For example, I have a struct of the form:
```
type form struct {
	// UserID userID for whom the query is being performed
	UserID string `form:"user_id" binding:"required,uuid"`
}
```
which represents the GET input parameters for my handler.
Before this fix, the generated YAML output says the parameter name is `userID`. The introspection looks for JSON tags on the struct but doesn't check for the presence of the `form` tag so it falls back to the field name.

By including the change, developers who explicitly set the `json` tag won't see any change in output, but those who provided only a `form` tag will set a result that will make the swagger file compatible with gin.

More generally, the `form` tag is not currently supported. This change will add support for it.

sdghchj:
What if a struct is shared by two routers, one use it as form request model, the other use it as response model?

sdghchj:
Is there a better way to support this? Such as parsing form tag into extensions with a special extension name and pick it when parsing request according to `@accept`, finally remove all the form extensions from all exported struct fields schemas.

pconstantinou:
Yes. An optimal solution might be to select the `json` vs. `form` tag based on the `@accept` tag to allow for the reuse of structs both as query parameters as well as JSON responses where you want (for reasons I can't imagine) to have inconsistent names. However, even in those cases, the likely scenario is that the `json` and `form` tags will match. In general, a tag is only provided when the default field name is not desired.

The only need for that enhancement is specificity when the tags are not consistent.

The current behavior on master is wrong for query parameters. It fails to generate correct parameter names for `GET` and `POST` parameter requests when `form` tags are used.

My change is an improvement in that JSON still takes priority when no other tag is used OR when the `json` tag is used. The  `form` tag will work if no other is tag specified. So it reduces the number of bugs without adding any new ones.

Without this change, how is one supposed to define an OPENAPI definition for query parameters using a struct?

pconstantinou:
This pull request is using the same approach for `query` tags.
https://github.com/swaggo/swag/pull/1251
@Ferrany1

ubogdan:
@pconstantinou As @sdghchj suggested, the struct can be shared between "accept" and "query", but after it gets parsed which models should we store into documentation the one handling JSON tags, the one with form tags, or the one with query tags?

rafamarqqqs:
Any updates on this? I'm also having some trouble using the uppercase acronym pattern naming in Go and requesting the camelcase style in the API with the `form` tag (e.g: `TaxID`  struct field name with the tag `form:"taxId"` becomes `taxID`, event with the default camelcase strategy).

sdghchj:
Anyway, this is a temporary solution.

sdghchj (approved):

PR #1428, opened 2022-12-20
docs: explain the need for doc comments when using swag fmt

morremeyer:
**Describe the PR**

This PR adds an explanation about the need for doc comments in addition to swagger comments when using `swag fmt`. 

**Relation issue**

Documentation for a rare issue that can occur since swag is now formatting with tabs since #1386.


ubogdan (approved):
LGTM

Issue #1430, opened 2022-12-23
'Missing required param comment parameters' on external library that not I need


vidorifaldy:
I'm trying to create swagger using Swaggo with command

`swag init --parseDependency`
(I have struct on shared library so I have to use this option)

unfortunately, I got this error
![Screenshot_20221222_050546](https://user-images.githubusercontent.com/49903427/209262916-49c8a8a2-7a3a-4b6b-af44-9e45f9d03a74.png)

Is there any option better than I copy the struct to my project?
How to use exclude the library?

thank you

Hope it goes well normally

swag version is v1.8.9
go version is v1.19

sdghchj:
I'm afraid no better way so far.

sdghchj:
Here I want to clarify what is `dependency`. 
Usually, we just want to referece models from external packages which are called `dependency`.
Is there any need to parse API operations in `dependency`? There is no need, I think.
@ubogdan What's your opinion?

vidorifaldy:
yes I need a model from external packages but got an error from another package that I don't need to parse

ubogdan:
> Here I want to clarify what is `dependency`. Usually, we just want to referece models from external packages which are called `dependency`. Is there any need to parse API operations in `dependency`? There is no need, I think. @ubogdan What's your opinion?

The only valid reason may be the API handlers are stored in a `dependecy` package. This will make the development harder and, we should deprecate/remove this in future versions.

vidorifaldy:
@sdghchj I got the insight from this 

<pre>https://github.com/swaggo/swag/issues/1014#issuecomment-1183215509</pre>

and then I change the command from this
`swag init --parseDependecy`
to this
` go mod vendor && swag init  --parseVendor --exclude ./vendor/github.com && rm -rf vendor`

It solved
but, I have a question, is it good or not?

sdghchj:
@vidorifaldy Could you please try the latest PR 1432?

vidorifaldy:
@sdghchj still get the same error

sdghchj:
> @sdghchj still get the same error

It seems impossible.

vidorifaldy:
@sdghchj thank you so much, It solved on v1.8.10

note: another issue after Merged PR 1432 [Issues 1451](https://github.com/swaggo/swag/issues/1451)

PR #1432, opened 2022-12-28
Do not parse operations in dependency

sdghchj:
**Describe the PR**
No longer parse operations in dependency, just parse models 

**Relation issue**
#1430


ubogdan (approved):

sdghchj:
After this PR, it is suggested that if you want to parse apis in some external packges，just explicitly include them with flag -d。
swag init -g main.go -d cmd/api,pkg/api

PR #1439, opened 2023-01-09
Enchancement: Report which property is triggering a parsing error

sakishrist:
Hi! Hope you could consider this change :)

**Describe the PR**
Add some more tracability to the errors generated by the parser. This can change this message:

```
**2023/01/09 17:33:42 ParseComment error in file /home/user/bla/bla/file.go :need array item type after array**
```

To a more descriptive:

```
**2023/01/09 17:33:42 ParseComment error in file /home/user/bla/bla/file.go :view.Events: Event: featureRequest: LocalizedItems: LocalizedItem: itemIds: need array item type after array
```

**Relation issue**
None

**Additional context**
What this solves is that with the first message we don't know where to look, but with the second message it explains which field the error ocured for as it might be far far away from the starting. The example above is the actual depth the parser was at, so it was throwing me off the scent of the origin of the error.

Tests are passing OK. The `errors.Is(...)` was added to find the wrapped error.

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -1029,7 +1029,7 @@ func (parser *Parser) getTypeSchema(typeName string, file *ast.File, ref bool) (
 			if err == ErrRecursiveParseStruct && ref {
 				return parser.getRefTypeSchema(typeSpecDef, schema), nil
 			}
-			return nil, err
+			return nil, fmt.Errorf("%s: %w", typeName, err)
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


Zhomart:
This looks good!

ubogdan:
Would you mind adding some unit tests to cover the changes?

sdghchj (approved):

Issue #1441, opened 2023-01-14
Remove hyphen from generated package name from docs.go


hwwwi:
**Is your feature request related to a problem? Please describe.**
Using the `--output` option with a hyphen included string generates an invalid package name on `docs.go`. 
Ex. `--output my-swag` will generate the `my-swag` package in `docs.go`. 
Using a hyphen in the go package name is invalid.
https://go.dev/ref/spec#Package_clause

**Describe the solution you'd like**
- We can easily replace hyphens to underscore for the package name.
- We can add an option to specify the package name.

PR #1442, opened 2023-01-14
feat: add packageName option for docs.go

hwwwi:
**Describe the PR**
swag is generating an invalid package name when the `--output` option is used with a hyphen.
Since using a hyphen in the package name is not allowed in go, replace it with an underscore.
Ex. `--output my-swag` will generate `package my_swag`

Also, I created a `packageName` option to specify the name of the package from generated `docs.go` file.
Ex. `swag init --packageName my_swag` will generate `package my_swag`.

**Relation issue**
fix https://github.com/swaggo/swag/issues/1441


ubogdan (approved):
LGTM

PR #1457, opened 2023-01-30
docs: remove unnecessary '$' from commands

Kukurijek:
**Describe the PR**
 remove unnecessary `$` from commands, because they are placed under the bash tag and can be directly copied, so the `$` is just unnecessary.

**Relation issue**
I did not create an issue for this because I think it is a little thing and there is no need to open an issue.

**Additional context**
Right now, if you copy e.g. `$ go install github.com/swaggo/swag/cmd/swag@latest` by clicking the button, you will get command that you cannot run, and without `$` sign, you can copy the command and run it directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ Swag converts Go annotations to Swagger Documentation 2.0. We've created a varie
 
 2. Download swag by using:
 ```sh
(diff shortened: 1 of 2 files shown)


ubogdan (approved):
LGTM

PR #1467, opened 2023-02-11
add support for big.Int

luonannet:
**Describe the PR**
add support for big.Int

**Relation issue**
none

**Additional context**

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -83,7 +83,7 @@ func TransToValidSchemeType(typeName string) string {
 		return INTEGER
 	case "uint32", "int32", "rune":
 		return INTEGER
-	case "uint64", "int64":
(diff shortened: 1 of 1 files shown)


schema.go
```diff
@@ -83,7 +83,7 @@ func TransToValidSchemeType(typeName string) string {
 		return INTEGER
 	case "uint32", "int32", "rune":
 		return INTEGER
-	case "uint64", "int64":
+	case "uint64", "int64", "big.Int":
```
ubogdan, line 86: What is the import is something like ?
```go
import integer "math/big"
```

On the other hand big.Int is not a primitive so this change is not valid.

ubogdan (changes requested):

Issue #1466, opened 2023-02-09
In versions greater than 1.8.4 --parseDependency crashes with indirect dependencies


ldrazic:
I have a gin api project in go, github.com/gin-gonic/gin uses the package github.com/pelletier/go-toml/v2 v2.0.6 that has:
const (
	// The default lowest and highest continuation byte.
	locb = 0b10000000
	hicb = 0b10111111)
	
this const in the utf8.go file

If I run `swag init --parseDependency=true -g docs/config_swagger.go` using the swaggo/swag v1.8.10

I get the next error: 

2023/02/09 16:59:51 warning: failed to evaluate const locb at /Users/leandrodrazic/go/pkg/mod/github.com/pelletier/go-toml/v2@v2.0.6/internal/characters/utf8.go:183:2, strconv.ParseUint: parsing "b10000000": invalid syntax
2023/02/09 16:59:51 warning: failed to evaluate const hicb at /Users/leandrodrazic/go/pkg/mod/github.com/pelletier/go-toml/v2@v2.0.6/internal/characters/utf8.go:184:2, strconv.ParseUint: parsing "b10111111": invalid syntax

But in version 1.8.4 this problem doesnt happen and the yaml is built without any error or warning

leimeng-go:
I have the same question

coolyrat:
I am currently using the `-q` to ignore the warning.

gfppower:
I have the same question

minQpple:
me too, I have to use v1.8.3

ZackaryWelch:
Having a similar issue on 1.8.10 with octal constants in the form "0o755" or the like. Tries to parse them as uint and fails.

minghsu0107:
Have the same issue too when using `v1.8.10`. Everything works fine after downgrading to `v1.8.4`.

sdghchj:
The same issue as #1584

PR #1475, opened 2023-02-23
adding binary to int conversion

ldrazic:
**Describe the PR**
adding binary to int conversion

**Relation issue**
https://github.com/swaggo/swag/issues/1466

**Additional context**
All the context is in the issue


package.go
```diff
@@ -94,6 +94,17 @@ func (pkg *PackageDefinitions) evaluateConstValue(file *ast.File, iota int, expr
...
 		switch valueExpr.Kind {
 		case token.INT:
+			// binary
+			if len(valueExpr.Value) > 2 && valueExpr.Value[0] == '0' && (valueExpr.Value[1] == 'b' || valueExpr.Value[1] == 'B') {
+				if x, err := strconv.ParseInt(valueExpr.Value[2:], 2, 64); err == nil {
+					return int(x), nil
+				} else if x, err := strconv.ParseUint(valueExpr.Value[2:], 2, 64); err == nil {
+					return x, nil
+				} else {
+					panic(err)
```
ubogdan, line 104: You shouldn't panic here. 
You should `return 0, nil`, and print the actual error.

ubogdan (changes requested):

ubogdan:
Please also add unit tests for this change.

package.go
ldrazic: I thought the same but looking at hexadecimal and octects conversions that are below my code, I tried to keep the same standard

ubogdan: Can you please fix the other "standard" :)

ubogdan:
@ldrazic, Any updates?

ldrazic:
@ubogdan I got stuck on the unit tests and left it there

sdghchj:
A same PR as #1593.

Issue #1490, opened 2023-03-06
Variable declaration with multiple variables of the same type in generic, causes type error


Bijan-Massoumi:
**Describe the bug**
When specifying the response type of an endpoint in the `@success` tag
This works
```
type (
	ResponseWithMetadata[T any, U any] struct {
		Error    *string `json:"error"`
		Data     T       `json:"data"`
		Metadata U       `json:"metadata,omitempty"`
	}
)
...
// @Success		200 {object} apiutils.Response[apisnapshot.SnapshotFramework]
```
this doesn't work:
```
type (
	ResponseWithMetadata[T, U any] struct {
		Error    *string `json:"error"`
		Data     T       `json:"data"`
		Metadata U       `json:"metadata,omitempty"`
	}
)
...
// @Success		200 {object} apiutils.Response[apisnapshot.SnapshotFramework]
```

it errors with this
```
2023/03/06 10:51:46 ParseComment error in file /Users/bmassoumi/work/governor-service/internal/api/handlers/governance/proposition.go :cannot find type definition: apisnapshot.SnapshotFramework
```

**Your swag version**
e.g. 1.8.10
**Your go version**
e.g. 1.19.0

**Desktop (please complete the following information):**
 - OS: macOS

hendrorahmat:
hi @sdghchj how can i use Model composition in Generic Struct, is that possible? for examples above, i used apiutils.Response[apisnapshot.SnapshotFramework]{data=string} and i got error :invalid type:

sdghchj:
> hi @sdghchj how can i use Model composition in Generic Struct, is that possible? for examples above, i used apiutils.Response[apisnapshot.SnapshotFramework]{data=string} and i got error :invalid type:

Not support yet.
Model composition is just an alternative to generics before go v1.18.
After go v1.18, is there any need to still use model composition?

PR #1506, opened 2023-03-19
fix generic type that has more than 2 params with only one constraint type

sdghchj:
**Describe the PR**
fix generic type that has more than 2 params with only one constraint type.
Before, we have only onsidered generic type format like [x any, y any], but missed  [x,y any].

**Relation issue**
#1490

### DIFF
--- a/generics.go
+++ b/generics.go
@@ -19,6 +19,11 @@ type genericTypeSpec struct {
 	Name       string
 }
 
+type formalParamType struct {
+	Name string
+	Type string
+}
+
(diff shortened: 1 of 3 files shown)


ubogdan:
Could you add some unit tests to cover this change?

ubogdan (approved):

sdghchj:
@ubogdan added

ubogdan (approved):

ubogdan:
@sdghchj let's get this merged and release a new version.

PR #1514, opened 2023-03-27
Included package infos related to the model

felagund18:
**Describe the PR**
The current generator does not include any package information. 
I had to create a restApi client that only imports the models without creating new ones.

**Additional context**
It only includes the params shown below

```json
"fiber.Args": {
    "type": "object",
    "x-pkg": {
        "full-path": "github.com/gofiber/fiber/v2.Args",
        "name": "Args",
        "package": "github.com/gofiber/fiber/v2",
        "type-name": "fiber.Args"
    }
},
```


Nerzal:
I guess this is something, that we should control using a CLI parameter. 
Not I guess not everyone wants package information contained in the documentation

Issue #1521, opened 2023-03-29
bug: @Param struct generator ignores enum array field


Jictyvoo:
**Bug Description**
Currently I have an enum that I want to use on my request, but it is an array on my form struct. When I try to generate the swagger docs, this field has its type generated, but it isn't added to request doc.

```go
type CustomEnum string

const (
    EnumA CustomEnum = "A"
    EnumB CustomEnum = "B"
    EnumC CustomEnum = "C"
)

type MyForm struct {
    Something bool `json:"something"`
    HereTheProblem []CustomEnum `json:"here_the_problem"`
}

// MyHandler godoc
// @Summary     Just a handler
// @Description Handler description
// @Tags        handler
// @ID          unique-id_oh_yeah
// @Produce     json
// @Param       data        query    MyForm true "filters"
// @Success     200         {object} SomeReturnStruct
// @Success     404         {object} dtos.Errors
// @Success     500         {object} dtos.Errors
// @Security    Token
// @Router      /myhandler [get]
func MyHandler(c *fiber.Ctx) error {
     // doesn't matters
    return nil
}

```

**To Reproduce**
- Use version [v1.8.12](https://github.com/swaggo/swag/releases/tag/v1.8.12)

**Expected behavior**
The swagger docs being generated with the given enum

**Swag version**
1.8.12

**Go version**
1.20.0

**Desktop:**
 - **OS:** Pop!_OS 22.04 LTS
 - **Browser:** Firefox
 - **Version:** 111.0.1 (64-bit)

chiptus:
@sdghchj as far as I can see, it generates the values, but not the `  x-enum-comments:` and `x-enum-varnames:`

PR #1523, opened 2023-03-31
transform struct fields of array enum type to query/formData params

sdghchj:
**Describe the PR**
 Transform struct fields of array enum type to query/formData params

**Relation issue**
#1521


ubogdan (approved):
Nice work!

PR #1543, opened 2023-04-10
fix(version): update version to 2

chiptus:
**Describe the PR**
update the version

I found out that running `swag --version` returns 1.8 which is weird

### DIFF
--- a/version.go
+++ b/version.go
@@ -1,4 +1,4 @@
 package swag
 
 // Version of swag.
-const Version = "v1.8.12"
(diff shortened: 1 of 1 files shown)


Nerzal:
That is caused by the fact, that the V2 implementation has not yet been merged to the master branch. 
It still resides on the "v2" branch, as it needs a lot of testing and i'm also currently on the "last" missing feature to get feature parity between master and v2 branch

chiptus:
Sorry, my mistake

PR #1552, opened 2023-04-14
V2: upgrade modulename to v2

Nerzal:
**Describe the PR**
In order to be able to import the package with a v2 version, we need to add /v2 to the module name

Else we get this error: 
> go: github.com/swaggo/swag/v2@v2.0.0-beta: go.mod has non-.../v2 module path "github.com/swaggo/swag" (and .../v2/go.mod does not exist) at revision v2.0.0-beta


Nerzal:
I'm not sure how to correctly test this change without merging and creating a new tag. 
So i'd suggest to merge this and i'll fix all issues then

Nerzal:
@ubogdan can u tell me why snyk is complaining?

Also is there a reason why snyk is not related to the swaggo organisation and instead tied to your personal account?

ubogdan:
@Nerzal I manage to link it to swaggo org.

ubogdan (approved):

Issue #904, opened 2021-03-24
Feature Request: Partially parse the dependencies.


CNLHC:
I found that when using the `--parseDependency` flag,  parsing would become very slow in large project.

I believe somebody else is suffering from the same problem, like #515 .

Maybe it is possible to refer the `go test` command. If i want to test a single case, i  can just fire
```
go test -timeout 30s -run ^<SomeRegExp>$ <ModuleName> -v
```

So to the `swag` tool, a flag like `include` will be helpful.  For example, the command below will only parse the `github.com/lib/pq` module instead of scanning all of the dependencies.

```
swag init --parseDependency --include "^github.com/lib/pq/$"
```

Do you think adding a filter mechanism to this feature is a good idea?

sdghchj:
--parseDepth 3

CNLHC:
> --parseDepth 3

Merely setting depth constraint can not solve the problem because the dependencies tree may be fat but shallow.

dskyu:
@CNLHC  can't agree more!

zskylark:
说得太对了，只要注解里引用了外部公共包里的类型定义就解析半天，就算是指定parseDepth也无济于事，经常解析20分钟以上。

SilverRainZ:
Does anyone work on this? This is a useful feature.

SilverRainZ:
I've submitted a [PR](#1582) to fix the issue. If you guys are interested, you can check out the corresponding branch.

PR #1582, opened 2023-05-08
feat: add --packagePrefix=P for only parse packages matched by prefix P

SilverRainZ:
This PR adds a `--packagePrefix=P` option for only parse packages matched by prefix P.
This is useful when our swagger annotations are located in dependency. In the meantime there are some 3rd-party dependency (that we have no control) has some malform swagger annotations, we can use this option to exactly parse the package we need.

Close https://github.com/swaggo/swag/issues/904

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -153,6 +153,10 @@ type Parser struct {
 	// excludes excludes dirs and files in SearchDir
 	excludes map[string]struct{}
 
+	// packagePrefix is a list of package path prefixes, packages that do not
+	// match any one of them will be excluded when searching.
+	packagePrefix []string
+
 	// tells parser to include only specific extension
 	parseExtension string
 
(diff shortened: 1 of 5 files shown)


SilverRainZ:
Hello @Nerzal, can you have a look?

sdghchj (dismissed):

sdghchj:
I wonder If one wants to parse more than one packages with different prefixes?

SilverRainZ:
Sure, P of `--packagePrefix=P` is a comma-separated string list, users can specify multiple prefixes if they want.

parser.go
```diff
@@ -275,7 +275,7 @@ func SetExcludedDirsAndFiles(excludes string) func(*Parser) {
 	}
 }
 
-// SetExcludedDirsAndFiles sets directories and files to be excluded when searching.
+// SetPackagePrefix sets directories and files to be excluded when searching.
```
SilverRainZ, line 278: This comment is copied by accident, maybe we should remove it.

sdghchj: It is required if the function is exportable.
Just correct the comment content,please.

SilverRainZ: done

SilverRainZ: I rebased to master by the way.

sdghchj (approved):

SilverRainZ:
Thanks for you reviewing. When this PR can be merged? :D

Issue #879, opened 2021-01-28
Add excluding file and placeholder file support


smiletrl:
**Is your feature request related to a problem? Please describe.**
The project might include api endpoints for different ports. For example, one port for mobile, and one port for desktop. In this case, `pkg/user` might include `pkg/user/api_mobile.go` and `pkg/user/api_desktop.go`, and each file includes the port's own endpoints.

The feature is to allow mobile version api doc to exclude all `api_desktop.go` file or `*_desktop.go`.

**Describe the solution you'd like**
Use command like `swag init --exclude api_desktop.go` or `swag init --exclude *_desktop.go` to exclude files `api_desktop.go` or `endpoints_desktop.go` for doc generating.

**Describe alternatives you've considered**
I was thinking the current option `--exclude` should achieve this already, but it is only excluding path now, but not specific file.

**Additional context**
Maybe the final goal could be ignoring files like what `.gitignore` ignores for git repository.

smiletrl:
It looks like I can't make a pull request against this repo, so here's a patch working for me  :)

```
diff --git a/cmd/swag/main.go b/cmd/swag/main.go
index 75c1756..5052540 100644
--- a/cmd/swag/main.go
+++ b/cmd/swag/main.go
@@ -40,7 +40,7 @@ var initFlags = []cli.Flag{
 	},
 	&cli.StringFlag{
 		Name:  excludeFlag,
-		Usage: "Exclude directories and files when searching, comma separated",
+		Usage: "Exclude directories and files (placeholder leading `*` is supported, e.g., `*_app.go`) when searching, comma separated",
 	},
 	&cli.StringFlag{
 		Name:    propertyStrategyFlag,
diff --git a/parser.go b/parser.go
index 4ca9bbf..43ce6cd 100644
--- a/parser.go
+++ b/parser.go
@@ -1440,6 +1440,17 @@ func (parser *Parser) Skip(path string, f os.FileInfo) error {
 				return filepath.SkipDir
 			}
 		}
+	} else {
+		if parser.excludes != nil {
+			if _, ok := parser.excludes[f.Name()]; ok { // exclude direct file name match
+				return filepath.SkipDir
+			}
+			for exclude := range parser.excludes {
+				if exclude[:1] == "*" && strings.Contains(f.Name(), exclude[1:]) { // exclude placeholder file `*some_file_name`
+					return filepath.SkipDir
+				}
+			}
+		}
 	}
 
 	return nil

```

[...]

PR #1585, opened 2023-05-18
exclude files too

Guillembonet:
**Describe the PR**
Allows excluding files as described in the `--exclude` flag help. Right now only directories can be excluded.

**Relation issue**
related to #879

**Additional context**
Files will be excluded in the `getAllGoFileInfo` function while directories continue to be excluded during the directory walk. This is why in the test I added there are 4 files instead of 3.


Guillembonet:
hi, I fixed the failing test (I think)

Guillembonet:
hey @sdghchj , can someone approve the workflow? thanks!

plusiv:
Hey folks @easonlin404 @ubogdan, could you please take a look to this related to #879 by any chance?

PR #1581, opened 2023-05-06
fix bug: enums of underscored number

sdghchj:
**Describe the PR**
support enums of underscored number

**Relation issue**
#1580

PR #1592, opened 2023-05-25
v2: fix bug: enums of underscored number (#1581)

Nerzal:
**Describe the PR**
Bring latest change from master to v2 branch

### DIFF
--- a/package.go
+++ b/package.go
@@ -5,6 +5,7 @@ import (
 	"go/token"
 	"reflect"
 	"strconv"
+	"strings"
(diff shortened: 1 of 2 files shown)


sdghchj:
> Would you mind adding some unit tests to cover this change ?

I think there is no much need to add a single unit test.

```
const underscored = 1_000_000
```

This line has covered this change. swag will try to parse all the consts at first.

sdghchj (approved):

PR #1621, opened 2023-07-06
fix required params parsing for routes with multiple paths and multiple params

### LINKED ISSUE #1267: Multiple path params with multiple paths leads to unnecessary required params
**Describe the bug**
When using multiple paths for a single endpoint, if any of the paths have a path param that is not contained in another path, all of the paths will contain that param and be marked as "required".

**To Reproduce**
Use the [example from the README](https://github.com/swaggo/swag#add-multiple-paths) to generate API specs. The generated spec will mark `group_id` as required for both paths.

**Expected behavior**
Paths that do not contain a path param will have that path param omitted in the generated spec.

**Your swag version**
v1.8.3

**Your go version**
go1.18.4

 - OS: Ubuntu 20.04.4

Phenix66:
**Describe the PR**
Fixes #1267 

**Relation issue**
[e.g. https://github.com/swaggo/swag/pull/118/files](https://github.com/swaggo/swag/issues/1267)

**Additional context**

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -977,7 +977,7 @@ func processRouterOperation(parser *Parser, operation *Operation) error {
 
 		if len(operation.RouterProperties) > 1 {
 			newOp := *operation
-			validParams := []spec.Parameter{}
+			var validParams []spec.Parameter
 			for _, param := range newOp.Operation.OperationProps.Parameters {
 				if param.In == "path" && !strings.Contains(routeProperties.Path, param.Name) {
 					// This path param is not actually contained in the path, skip adding it to the final params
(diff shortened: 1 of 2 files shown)


parser.go
```diff
@@ -975,7 +975,21 @@ func processRouterOperation(parser *Parser, operation *Operation) error {
 			parser.debug.Printf("warning: %s\n", err)
 		}
 
-		*op = &operation.Operation
+		if len(operation.RouterProperties) > 1 {
+			newOp := *operation
+			validParams := []spec.Parameter{}
```
ubogdan, line 980: Please use `var validParams []spec.Parameter` in order to avoid mem allocations.

ubogdan (changes requested):

parser.go
Phenix66: Updated

ubogdan (approved):

ubogdan:
@Phenix66 Thanks for your contribution

Issue #1623, opened 2023-07-12
--tags "!this" does not work


rsmarples:
**Describe the bug**
Tag filtering assumes at least one positive match

**Expected behavior**
I would expect all API's to not have the tags parsed.

**Your swag version**
1.8.12

**Your go version**
1.19.7

**Desktop (please complete the following information):**
 - OS: DragonflyBSD
 - Browser: FireFox
 - Version: 111

PR #1624, opened 2023-07-12
parser: if all tags negate return true on no hits

rsmarples:
**Describe the PR**
Fix tag parsing in the case when all tags are negation.

**Relation issue**
https://github.com/swaggo/swag/issues/1623

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -867,18 +867,29 @@ func getTagsFromComment(comment string) (tags []string) {
 }
 
 func (parser *Parser) matchTags(comments []*ast.Comment) (match bool) {
-	if len(parser.tags) != 0 {
-		for _, comment := range comments {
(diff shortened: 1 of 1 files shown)


sdghchj (approved):

parser.go
```diff
@@ -867,18 +867,29 @@ func getTagsFromComment(comment string) (tags []string) {
...
+			}
+			if _, has := parser.tags[tag]; has {
+				match = true // keep iterating as it may contain a tag that is excluded
+			}
+		}
+	}
+
+	if !match {
+		// If all tags are negation then we should return true
+		for key, _ := range parser.tags {
```
sdghchj, line 888: for key := range parser.tags

sdghchj (changes requested):

sdghchj (approved):

Issue #1608, opened 2023-06-20
String Enum in body Param generating incorrect schema


SulanthaM:
**Describe the bug**
Incorrect spec is generated when a body param defined with string Enum type. 

**To Reproduce**
Steps to reproduce the behavior: Definition
`@Param modify-org-type body string false "type" Enums(PHARMA, ACADEMIC, VENDOR, INDUSTRY, PARTNER)`

**Result**
```
                 {
                        "enum": [
                            "PHARMA",
                            "ACADEMIC",
                            "VENDOR",
                            "INDUSTRY",
                            "PARTNER"
                        ],
                        "description": "type",
                        "name": "modify-org-type",
                        "in": "body",
                        "schema": {
                            "type": "string"
                        }
                    }
```

**Expected behavior**
Correct/Valid swagger schema

```
                 {
                  
                        "description": "type",
                        "name": "modify-org-type",
                        "in": "body",
                        "schema": {
                            "type": "string",
                            "enum": [
                              "PHARMA",
                              "ACADEMIC",
                              "VENDOR",
                              "INDUSTRY",
                              "PARTNER"
                              ]
                        }
                    }
```

**Your swag version**
e.g. 1.16.0

**Your go version**
e.g. 1.19.0

ubogdan:
Merged.

PR #1625, opened 2023-07-17
fix: enums with `paramType`=body got parsed incorrectly

hohobilly:
**Describe the PR**
Attempt to fix the issue #1608

**Relation issue**
https://github.com/swaggo/swag/issues/1608

**Additional context**
Adding a test case with the commentLine expression similar to the one suggested in the issue.


ubogdan (approved):

ubogdan:
@hohobilly Thanks for your contribution!

Issue #1662, opened 2023-08-28
swag init: panic: recursively parsing struct


jeffwalter-rum:
**Describe the bug**
I have the following data strutures:
```
import "github.com/my/local/openrtb"

type Response struct {
	RTB        *openrtb.BidResponse
}
```
and the following package (github.com/my/local/openrtb):
```
package openrtb

import "github.com/bsm/openrtb"

type BidResponse = openrtb.BidResponse
```

I get the following panic:
```
2023/08/28 12:41:11 Generating openrtb.BidResponse
2023/08/28 12:41:11 Skipping 'openrtb.BidResponse', recursion detected.
2023/08/28 12:41:11 Error parsing type definition 'openrtb.BidResponse': recursively parsing struct
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x100c769fc]

goroutine 1 [running]:
github.com/swaggo/swag.(*Parser).getRefTypeSchema(0x140000afd40, 0x14000c3e0a0, 0x0)
        /Users/user/go/pkg/mod/github.com/swaggo/swag@v1.16.1/parser.go:1073 +0x7c
github.com/swaggo/swag.(*Parser).getTypeSchema(0x140000afd40, {0x14000c77380, 0x13}, 0xd?, 0x1)
        /Users/user/go/pkg/mod/github.com/swaggo/swag@v1.16.1/parser.go:1052 +0xac8
github.com/swaggo/swag.(*Parser).parseStructField(0x140000afd40, 0x2d?, 0x1400067b5c0)
        /Users/user/go/pkg/mod/github.com/swaggo/swag@v1.16.1/parser.go:1373 +0x33c
github.com/swaggo/swag.(*Parser).parseStruct(0x14000814160?, 0x10?, 0x1400035bf20)
        /Users/user/go/pkg/mod/github.com/swaggo/swag@v1.16.1/parser.go:1286 +0xac
github.com/swaggo/swag.(*Parser).parseTypeExpr(0x14000436750?, 0x100c9fcf4?, {0x100e636d8?, 0x14000671218?}, 0x88?)
        /Users/user/go/pkg/mod/github.com/swaggo/swag@v1.16.1/parser.go:1239 +0x70
github.com/swaggo/swag.(*Parser).ParseDefinition(0x140000afd40, 0x14000c3fa40)
[...]

PR #1666, opened 2023-09-06
fix: search for definitions from dependent packages first

sdghchj:
**Describe the PR**
If a type is an alias type from a dependent package and they share the same short package name and type name, the dependent type will be hidden. So this PR will search for type definitions from dependent packages first.

**Relation issue**
#1662


ubogdan (approved):

ubogdan:
Not sure about the change impact. You should write to cover this change. Approved.

PR #1704, opened 2023-11-15
Optimize docker container usage

ngehrsitz:
Currently the docker container has to be invoked with
```sh
docker run --rm -v $(pwd):/code -w /code ghcr.io/swaggo/swag:v1.16.2 /root/swag
```
which means users have to know that the binary is placed under `/root/swag`. If they don´t know this and just try
```sh
docker run --rm -v $(pwd):/code ghcr.io/swaggo/swag:v1.16.2
```
they will get the error
```sh
docker: Error response from daemon: No command specified.
```
if they then try
```sh
docker run --rm -v $(pwd):/code ghcr.io/swaggo/swag:v1.16.2 swag
```
it won´t work either
```sh
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "swago": executable file not found in $PATH
: unknown.
```
This PR adds an `ENTRYPOINT` meaning the first command will just work. It also moves the binary to `/bin` since having the binary in the same directory as the default `WORKDIR` is suboptimal since users need to mount their code into the container. If the binary is not stored in the `WORKDIR`, then users can simply mount their code to that directory without having to specify a different one with `-w`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,7 @@ RUN CGO_ENABLED=0 GOOS=linux go build -v -a -installsuffix cgo -o swag cmd/swag/
 ######## Start a new stage from scratch #######
 FROM scratch
 
-WORKDIR /root/
+WORKDIR /code/
 
 # Copy the Pre-built binary file from the previous stage
 COPY --from=builder /app/swag /bin/swag
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # Dockerfile References: https://docs.docker.com/engine/reference/builder/
 
 # Start from the latest golang base image
-FROM golang:1.18.3-alpine as builder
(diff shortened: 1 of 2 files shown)


ubogdan:
Would you mind adding some docker usage tips in the Readme.md?

ngehrsitz:
> Would you mind adding some docker usage tips in the Readme.md?

Did you notice the section I added there? I am not sure what else would be relevant?

ubogdan (approved):

ubogdan:
@ngehrsitz Thanks for your contribution!

ngehrsitz:
@ubogdan Please create a new release, so that `ghcr.io/swaggo/swag:latest` will actually work as described.

PR #1722, opened 2023-12-17
Change dockerfile base image

mjudeikis:
**Describe the PR**
When building more complicated apps go is a mandatory dependency In the image due to https://github.com/swaggo/swag/blob/6cdaaf5c77457e82d9e0f8fccd303fefb8dc8072/golist.go#L14-L15
So we need to use image with go.

Otherwise, it fails with error `go not in path`

In addition, can somebody confirm GitHub image latest works fine? We had to rebuild manually as one in GitHub looks corrupted.


sdghchj:
Hello, rebase your branch on the latest master, please.

PR #1715, opened 2023-12-11
update modules

jvilhuber:
**Describe the PR**

Update modules (go get -u ./...; go mod tidy; make) to make security scanners happy.

**Relation issue**

**Additional context**

PR #1723, opened 2023-12-18
Drop support for go v1.17.x

sdghchj:
**Describe the PR**
Remove go version v1.17.x

**Relation issue**
#1715 
#1722


sdghchj:
@ubogdan  Do you know how to fix it ?
I wonder Why it gets the latest go-openapi/spec version while v.0.20.4 is declared in go.mod？

ubogdan:
Well, I guess the "deps" target from the Makefile is not needed anymore. Also, we use GOPATH in the workflow named ci.yaml (which is not necessary anymore).

sdghchj:
> Well, I guess the "deps" target from the Makefile is not needed anymore. Also, we use GOPATH in the workflow named ci.yaml (which is not necessary anymore).

OK.
I have fixed it with `go mod tidy`.

ubogdan (approved):

Issue #1714, opened 2023-12-05
How to reponse a binary data type?


AmazingPangWei:
As title described, I've tried this:

```
// @Summary MyBinary
// @Produce octet-stream
// @Success 200 {string} binary "mybinary"
// @Router /myapis/myBinary [post]
```

and `swagger.json` like this:

``` json
        "/myapis/myBinary": {
            "post": {
                "produces": [
                    "application/octet-stream"
                ],
                "summary": "MyBinary",
                "responses": {
                    "200": {
                        "description": "mybinary",
                        "schema": {
                            "type": "string"
                        }
                    }
                }
            }
        }
```

As far as I know, according to the [documentation of OpenAPI](https://swagger.io/specification/v2/#dataTypeFormat),  `schema.format` should be `binary`, but I don't know how to specify it.

PR #1754, opened 2024-02-06
Support primitive type with format in response

sdghchj:
**Describe the PR**
Support primitive type with format in response

**Relation issue**
#1714

### DIFF
--- a/operation_test.go
+++ b/operation_test.go
@@ -244,6 +244,33 @@ func TestOperation_ParseResponseWithDefault(t *testing.T) {
 	assert.Equal(t, "A response", operation.Responses.StatusCodeResponses[200].Description)
 }
 
+func TestParseResponsePrimitiveTypeWithFormat(t *testing.T) {
+	t.Parallel()
+
+	comment := `@Success 200 {integer} int16 "response with format"`
+	operation := NewOperation(nil)
+	err := operation.ParseComment(comment, nil)
+	assert.Error(t, err)
+
+	comment = `@Success 200 {integer} int32 "response with format"`
+	operation = NewOperation(nil)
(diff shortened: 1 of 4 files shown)


schema.go
```diff
@@ -36,6 +36,12 @@ const (
 	IgnoreNameOverridePrefix = '$'
 )
 
+var validPrimitiveFormat = map[string]map[string]bool{
+	INTEGER: {"int32": true, "int64": true},
+	NUMBER:  {"float": true, "double": true},
+	STRING:  {"byte": true, "binary": true, "date": true, "date-time": true, "password": true},
+}
+
```
ubogdan, line 44: Why don't use the reverse way
```go
var validPrimitiveFormat = map[string]string{
   "int32": "INTEGER"
   "int64": "INTEGER"
   "float": "NUMBER"
    ....
}
```
and the condition from operations.go
```go
if !validPrimitiveFormat[refType] != schemaType {
```

operation_test.go
```diff
@@ -244,6 +244,33 @@ func TestOperation_ParseResponseWithDefault(t *testing.T) {
...
 
+func TestParseResponsePrimitiveTypeWithFormat(t *testing.T) {
+	t.Parallel()
+
+	comment := `@Success 200 {integer} int16 "response with format"`
+	operation := NewOperation(nil)
+	err := operation.ParseComment(comment, nil)
+	assert.Error(t, err)
+
+	comment = `@Success 200 {integer} int32 "response with format"`
```
ubogdan, line 255: This is a stupid corner case , Can you show me an example how you send and receive an int32 over an HTTP request ? :) 
Usually the response is a JSON object, not a string value :flashlight:

PR #1777, opened 2024-03-20
feat: ParseComment error to contain the comment

stokito:
When a comment has an error then we print only a file name and the error text but not the comment that caused an error. To make it easier to understand print it.

**Context:**
I spent a lot of time while investigating the error message:

    Generating dto.DeleteRequest
    Key is not supported paramType

It turned out that the DTO was fine but a problem was in another place with a comment:

    // @Param User Key body dto.UserRequest true "Event"

In the comment instead of `UserKey` it was written `User Key` and the Key was threatened as a `paramType`. I wasn't able to figure out which key they are talking about.

So the change should significantly improve troubleshooting which is important.

### DIFF
--- a/operation.go
+++ b/operation.go
@@ -386,7 +386,7 @@ func (operation *Operation) ParseParamComment(commentLine string, astFile *ast.F
 			param.Schema = schema
 		}
 	default:
-		return fmt.Errorf("%s is not supported paramType", paramType)
+		return fmt.Errorf("not supported paramType: %s", paramType)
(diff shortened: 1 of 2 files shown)


sdghchj (approved):

ubogdan:
@stokito can you please fix the conflicts.

ubogdan (approved):
LGTM

ubogdan:
@stokito Thanks for your contribution!

Issue #1839, opened 2024-07-04
Bad operation when there is a newline character between annotations


mariusz-4chain:
**Describe the bug**
Bad operation when there is a newline character between annotations

**To Reproduce**
1. Add a couple of annotations describing the endpoint
2. Enter a newline character between the annotations e.g.:
// @Router		/v1/transaction/record [post]

// @Security	x-auth-xpub
4. Documentation won't be generated or @Tags information is ignored and endpoint is assigned to default tag

**Expected behavior**
If the newline character is not allowed the plugin should throw an error and not silently ignore it.

**Screenshots**
![image](https://github.com/swaggo/swag/assets/166371395/8a9ac38b-2546-418b-8300-7d260abbc54a)

**Your swag version**
e.g. 1.16.3

**Your go version**
 go1.22.4 linux/amd64

**Desktop (please complete the following information):**
 - OS: Linux Mint 21.3
 - Browser: Chrome
 - 126

PR #1846, opened 2024-07-19
fixed operation when there is a newline character between annotations

mahiworld:
This change resolves issue #1839 by adding the parse newline annotation and throw an error: newline detected between annotations.
Issue: https://github.com/swaggo/swag/issues/1839

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -1,6 +1,7 @@
 package swag
 
 import (
+	"bufio"
 	"context"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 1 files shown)


Danli741:
I have tested manually (with example/celler)
With new line ("without //") still does not work. 
image

ubogdan:
Cool! What about unit tests?

ubogdan:
@mahiworld in case you missed my reply. 
 If you want to get this merged you need to provide unit test to cover the code you added.

mahiworld:
@ubogdan Added unit test for the change & there's some merge conflict, resolved that as well.

PR #1848, opened 2024-07-22
Add support for generating pointers as x-nullable

Thiht:
**Describe the PR**

With this change, pointers inside a struct are automatically generated with the x-nullable extension

New CLI flag: `--nullablePointers`

- Pointers in structs are generated with `x-nullable: true`
- If the schema contains a single `$ref`, like this:

```yml
my_field:
  $ref: "#/definitions/api.Child"
  x-nullable: true
```

it will be generated like this instead:

```yml
my_field:
  allOf:
    $ref: "#/definitions/api.Child"
  x-nullable: true
```

This ensures the result can then be converted to OpenAPI 3.0 without losing the `nullable` field (using: https://converter.swagger.io). See: https://stackoverflow.com/a/48114924

I'm willing to make some improvements if you have any suggestions, I'm not sure I made the change correctly as it just fits my use case for now :) 

**Relation issue**
none

**Additional context**

I want to generate x-nullable on my pointers but I don't want to annotate all my fields with `extensions:x-nullable`. I think it makes sense to allow generating the x-nullable by default for all pointers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # swag
 
-🌍 *[English](README.md) ∙ [简体中文](README_zh-CN.md) ∙ [Português](README_pt.md)*
+🌍 _[English](README.md) ∙ [简体中文](README_zh-CN.md) ∙ [Português](README_pt.md)_
 
 <img align="right" width="180px" src="https://raw.githubusercontent.com/swaggo/swag/master/assets/swaggo.png">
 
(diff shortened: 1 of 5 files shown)


ubogdan:
I recommend you to open an issue before implementing anything.

Thiht:
@ubogdan haha sorry about that, I actually forked to add the feature for my team (we currently use the fork) and decided to try to upstream afterwards

I can still open an issue if you prefer and want to discuss some things, and if you have any feedback I'll be glad to make the changes

ubogdan:
I don't have any issue adding this improvement to the code; right now, I don't see its usage.

PR #1894, opened 2024-10-02
Flags to parse internal and dependency package

bytesByHarsh:
**Remove cannot find type definition error**
```go
package model

type PaginatedListResp[T any] struct {
	Data         []T  `json:"data"`
	TotalCount   int  `json:"total_count"`
	HasMore      bool `json:"has_more"`
	Page         int  `json:"page"`
	ItemsPerPage int  `json:"items_per_page"`
}

type User struct {
	ID          uuid.UUID `json:"id"`
	Name        string    `json:"name"`
	PhoneNum    string    `json:"phone_number"`
	Email       string    `json:"email"`
	Username    string    `json:"username"`
}
```

```go
package api
.
.
.
//	@Success	200	{object}	models.PaginatedListResp[models.User]
```

**Additional context**
If the model is part of an external dependency or internal package then flags mentioned in the PR can be used but were not mentioned in the generics section.

In some cases simple `import _ github.com/xxx/models` also works but if it does not then adding either `--parseDependency` or `--parseInternal` helps in solving the issue based on requirement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -947,6 +947,17 @@ swag init -g http/api.go -td "[[,]]"
 ```
 The new delimiter is a string with the format "`<left delimiter>`,`<right delimiter>`".
 
+### Parse Internal and Dependency Packages
+
+If the struct is defined in a dependency package, use `--parseDependency`.
+
+If the struct is defined in your main project, use `--parseInternal`.
+
+if you want to include both internal and from dependencies use both flags 
+```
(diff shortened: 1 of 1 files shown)


ubogdan (approved):

ubogdan:
@bytesByHarsh Thanks for your contribution!

PR #1903, opened 2024-10-17
merge master to v2

ubogdan:
**Describe the PR**
Sync changes from master with v2 branch. 

**Relation issue**

**Additional context**
Neet to resolve conflicts. (wip)


ubogdan:
@sdghchj  can you please give a hand here, we have 2 unit tests failing ...

sdghchj:
My pleasuere, however what can I do for you?
I have no developing experience about v2 mechanism.
I just doubt v2 was wrongly used.
I see the v2 branch is trying to support OpenAPI specification both V2 and V3, that means it is trying to be compatible with the master branch. But v2 is usually used when the compatibility is broken.
I just tried removing the `v2` from the module path and all the import package path in the branch `v2-merge` you created a few hours ago ,  then I found the test `TestParseGenericsFunctionScoped` passed.
So my opinion is remove drop v2. 
@Nerzal  Here needs your advice.

ubogdan:
@sdghchj No worries about the experience with the v2 branch.

 We changed the import path so v2 can be used/tested until the development is ready. The reason is obvious: many community members want OAS3, and a smaller one is open to testing and contributing.
  Also, I avoided merging so many changes into the master branch and breaking the v1.x version. 

The main reason we are doing this is to keep OAS 2 fixes in sync with v2, This means we can adapt the test import path to "..v2/,,,' . 

Keeping both OAS2 and OAS 3 in v2 will be more like transition path, When we consider the development is done we can warn users that we accept only security patches for "OAS2' and we will drop it in v3.

@sdghchj @easonlin404 @Nerzal  what is your suggestion here: Should we drop OAS 2 completely in v2? 











 -

ubogdan:
@sdghchj Thanks for the tip. Unit tests are fixed now.

sdghchj (approved):

njacob1001:
This merge may have caused a similar issue in v1
https://github.com/swaggo/swag/issues/1909

Issue #1919, opened 2024-10-29
 swag init generates documentation too slowly


luoxianjie:
**Is your feature request related to a problem? Please describe.**
As our projects become larger and larger, swag init will be very slow to generate documents, usually taking more than half an hour, which is very painful.

**Describe the solution you'd like**
Looking at the source code, we found that it takes a lot of time to search for dependency packages, because all dependencies will be searched repeatedly. It is not as good as both B and C depending on A. When building B and C, A will be searched repeatedly. We add the FindPackage method. After add the cache, I found that the speed increased by nearly 10 times. The code is as follows

```
conf := loader.Config{
  ParserMode:  goparser.ParseComments,
  Cwd:         cwd,
  FindPackage: FindPackage,
}
```

```
var packageCache = make(map[string]*build.Package)
var cacheMutex sync.RWMutex
var packageLocks sync.Map

func FindPackage(ctxt *build.Context, importPath, fromDir string, mode build.ImportMode) (*build.Package, error) {
	cacheMutex.RLock()
	pkg, found := packageCache[importPath]
	cacheMutex.RUnlock()

	if found {
		return pkg, nil
	}

	lockInterface, _ := packageLocks.LoadOrStore(importPath, &sync.Mutex{})
	lock := lockInterface.(*sync.Mutex)

	lock.Lock()
	defer lock.Unlock() 

	cacheMutex.RLock()
	pkg, found = packageCache[importPath]
	cacheMutex.RUnlock()

	if found {
		return pkg, nil
	}

	pkg, err := ctxt.Import(importPath, fromDir, mode)
	if err != nil {
		return nil, err
	}

	cacheMutex.Lock()
	packageCache[importPath] = pkg
	cacheMutex.Unlock()

	return pkg, nil
}
```

sdghchj:
感谢，但是我觉得在代码注释中用中文不太合适，毕竟这是面向所有开源开发者all over the world。

PR #1920, opened 2024-10-29
findPackage add cache

luoxianjie:
**Describe the PR**
As our projects become larger and larger, swag init will be very slow to generate documents, usually taking more than half an hour, which is very painful.

**Relation issue**
e.g. https://github.com/swaggo/swag/issues/1919

### DIFF
--- a/packages.go
+++ b/packages.go
@@ -3,15 +3,16 @@ package swag
 import (
 	"fmt"
 	"go/ast"
+	"go/build"
 	goparser "go/parser"
 	"go/token"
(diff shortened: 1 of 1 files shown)


sdghchj:
感谢，但是我觉得在代码注释中用中文不太合适，毕竟这是面向所有开源开发者all over the world。看能不能改一下？

easonlin404:
+1 to @sdghchj, and it would be great if we can have uni test and benchmark.

Issue #1940, opened 2024-12-10
global override int64 generated as integer


yabinma:
**Describe the bug**
When defining a global override and specify the target format as int64,  the type generated is integer, but without format:int64. So, it is treated as int32.

**To Reproduce**
1. Add .swaggo and add override,
`replace time.Duration int64`
2. run command to generate swagger.yaml file
`swag i --md docs/ --parseDependency  --parseInternal --outputTypes yaml  --parseDepth 1 --output .`
3. check the swagger.yaml file and  the type is integer, without format:int64

**Expected behavior**
The attribute generated should be with a format:int64

**Screenshots**
![image](https://github.com/user-attachments/assets/5580222b-162a-4b2f-a4bd-39115d6aca3e)

**Your swag version**
swag version v1.16.4

**Your go version**
go1.23.1

**Desktop (please complete the following information):**
 - OS:  Linux,ubuntu 22.04
 - Browser: [e.g. chrome, safari]
 - Version: [e.g. 22]

**Additional context**

PR #1944, opened 2024-12-11
Transfer golang type to swagger type with format reserved

sdghchj:
**Describe the PR**
Transfer golang type to swagger type with format reserved

**Relation issue**
#1940


ubogdan (approved):
LGTM

PR #1946, opened 2024-12-17
Send the version to stderr

greg0ire:
Until https://tip.golang.org/doc/modules/managing-dependencies#tools is
available to me, I have a shell script that detects the version of swag
by invoking swag --version and parsing its output. With v2, the output
looks like this

Swag version:  v2.0.0
swag version v2.0.0

While it might be useful to have the swag version always printed for
troubleshooting purposes, I think it should be printed to stderr so as
not to mess up output that is supposed to be parsed.

This change will result in the first line to be sent to stderr, making
it backward-compatible with the previous output of swag --version while
still retaining the improvement of having it printed for every command.


greg0ire:
The whole script looks as follows (it's actually a just recipe):

```just
tool: ci-tool
	#!/usr/bin/env bash
	set -euo pipefail
	# golang.org/x/tools ...
	go install golang.org/x/tools/...@latest
	echo "mockery installation ..."
	go install github.com/vektra/mockery/v2/...@latest
	echo "mockery: " `mockery --version`
	# Note that the `+x` part in the comparison of the two variables is used to handle cases where one of the variables is
	# empty (i.e., the `+x` converts an empty string to the string 'x').
	VERSION=$(swag --version 2>/dev/null | sed -rn "s/.* (v[0-9]+.[0-9]+.[0-9]+)$/\1/p");
	if [ $VERSION+"x" != {{SWAGGO_VERSION}}+"x" ]; then
		echo "swaggo installation ({{SWAGGO_VERSION}})";
		go install github.com/swaggo/swag/v2/cmd/swag@{{SWAGGO_VERSION}};
		swag --version;
	fi
```

Issue #1041, opened 2021-11-01
Example of nested array in response


ilaripih:
**Is your feature request related to a problem? Please describe.**
In a struct tag it's possible to use a nested array as **swaggertype**. For example:

```
type MyStruct {
  Data [][]int `json:"data" swaggertype:"array,array,number"`
}
```

This works fine, but the **example** tag doesn't work because Swag will simply split the example value into parts (1D array) with a comma separator here: https://github.com/swaggo/swag/blob/master/parser.go#L1487

**Describe the solution you'd like**
If there is no easy workaround without code changes in Swag (I may have missed something obvious), there is a pretty straightforward way to solve this: if **schemaType** is **ARRAY** and **arrayType** is also **ARRAY**, assume that **exampleValue** is a JSON-encoded string, decode it and return the decoded result. So the struct definition could look like this:

```
type MyStruct {
  Data [][]int `json:"data" swaggertype:"array,array,number" example:"[[0,1],[2,3]]"`
}
```

**Describe alternatives you've considered**
Maybe there could be another way to encode nested arrays (i.e. not JSON) in example tags, or a new schema type, e.g. "jsonarray"?

ubogdan:
@ilaripih `swaggertype` can't describe a complex object.

ilaripih:
But nested types like "array,array,number" are accepted and correctly parsed by Swag. The only thing that's missing is parsing the example value correctly. I can write a pull request that implements this and hopefully communicates my idea more clearly.

ubogdan:
@ilaripih you're welcome.

khalidiya:
how i need to describe examples of array of arrays?
for example [][]float64

PR #1956, opened 2024-12-30
Support array of array examples

ayca-aydogan:
**Describe the PR**
Add a nested array parser for an array of array examples. To achieve this, the example string is parsed into a tree and then the tree structure is converted to an array of arrays. "{" and "}" are chosen to specify the nested array beginning and end but this can easily be changed. An example of a nested array example is:```{{one,two},{three,four}}```. This implementation is backward compatible with arrays. The example could be "{one,two}" as well as "one,two"

**Relation issue**
https://github.com/swaggo/swag/issues/1041

**Additional context**
For an extensive fix, [depth in ComplementSchema](https://github.com/swaggo/swag/blob/94ff0fcc35857862b6c1f2bac2063e84522b6b50/field_parser.go#L244) should be a variable instead of constant 2.

### DIFF
--- a/field_parser_test.go
+++ b/field_parser_test.go
@@ -43,6 +43,36 @@ func TestDefaultFieldParser(t *testing.T) {
 			}},
 		).ComplementSchema(&schema)
 		assert.Error(t, err)
+
+		schema = spec.Schema{}
+		schema.Type = []string{"array"}
+		schema.Items = &spec.SchemaOrArray{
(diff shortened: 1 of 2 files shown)


qktrn:
When is this gonna be merged?

podocarp:
I think this is a fairly common thing and should  be looked at. Think matrices, csv, etc., nested arrays are not that uncommon.

PR #1960, opened 2025-01-09
Allow parenthesis in examples

mloesch:
**Describe the PR**
Allow parentheses in `Example` and `SchemaExample`, e.g. `Example(True feelings (maybe))`

**Additional context**
The implementation assumes that the `[Schema]Example` tag will always be the last tag in a comment string.

### DIFF
--- a/operation.go
+++ b/operation.go
@@ -426,23 +426,23 @@ const (
 
 var regexAttributes = map[string]*regexp.Regexp{
 	// for Enums(A, B)
-	enumsTag: regexp.MustCompile(`(?i)\s+enums\(.*\)`),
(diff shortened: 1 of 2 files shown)


sdghchj:
Your PR needs the `example` to be at the end.
Consider a case:
```
@param ……            Example(True feelings (maybe))  enums(1,2,3)
```
`True feelings (maybe))  enums(1,2,3` will be the example content.

mloesch:
I see, then I fear the solution for this problem is not trivial, as we would probably need to identify the balanced parenthesis - not sure if this is even possible with a regex - closing the PR for now.

PR #1975, opened 2025-02-10
deps: bump sigs.k8s.io/yaml and use gopkg.in/yaml.v3

nikpivkin:
Small dependency update. `gopkg.in/yaml.v3` has many fixes compared to v2. `sigs.k8s.io/yaml` v1.4.0 no longer depends on `gopkg.in/yaml`.


devhaozi:
This can be closed due to #1987.

PR #1976, opened 2025-02-10
fix: parse security definition key for Bearer auth

nikpivkin:
This PR fixes an issue that caused the tool to always generate a security definition with the `bearerauth` key, regardless of the user-specified value.


amaneshi:
This should be fixed already with https://github.com/swaggo/swag/pull/2027 so can be closed, imho

nikpivkin:
Already fixed https://github.com/swaggo/swag/pull/2027

PR #1978, opened 2025-02-13
[feature]: exclude api from UI with specified annotation

rvarunrathod:
**Describe the PR**
Add new flag ExcludeFromUI, This will remove API and unrelated definitions from only doc.go file so it won't be visible on Swagger UI.

**Additional context**
Hi Team, For long time I wanted this feature which just remove API from UI, so users can't access it from UI but it will be there in swagger.json/yaml so I can share this with other developers.

Usage:

```
// ShowAccount godoc
//
//	@Summary		Show an account
//	@Description	get string by ID
//	@Tags			accounts
//	@Accept			json
//	@Produce		json
//	@Param			id	path		int	true	"Account ID"
//	@Success		200	{object}	model.Account
//	@Failure		500	{object}	httputil.HTTPError
//	@Router			/accounts/{id} [get]
//      @x-hidden               true
```

```shell
go run cmd/swag/main.go init -d example/celler --excludeFromUI hidden
```
It will remove the API from UI which have this hidden annotation.

### DIFF
--- a/cmd/swag/main.go
+++ b/cmd/swag/main.go
@@ -43,6 +43,7 @@ const (
 	packagePrefixFlag        = "packagePrefix"
 	stateFlag                = "state"
 	parseFuncBodyFlag        = "parseFuncBody"
+	excludeFromUIFlag        = "excludeFromUI"
 )
 
 var initFlags = []cli.Flag{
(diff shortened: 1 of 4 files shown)


sdghchj:
As I know,  there is already a similar feature which uses `Tags`  to filter.

rvarunrathod:
Tags will remove this API definitions from swagger.json also, here use case is only remove it from UI code.

sdghchj:
> Tags will remove this API definitions from swagger.json also, here use case is only remove it from UI code.

Sorry, I am confused why this feature needed?

rvarunrathod:
Hi @sdghchj, For long time I wanted this feature which just remove API from UI, so users can't access it from UI, but it will be there in swagger.json/yaml so I can share this with other developers.

PR #2015, opened 2025-04-30
fix: update field parser v3 test case

taba2424:
**Describe the PR**
Fix type mismatch error in field parser test cases. 

**Relation issue**

**Additional context**


sdghchj:
conflict

Issue #2040, opened 2025-07-14
Treat json:omitempty as required: false when generating swagger output


andyatmiami:
**Is your feature request related to a problem? Please describe.**
When working on a large codebase with many complex, nested models that get surfaced in the Swagger specification, it can be challenging to ensure all field tags supply the appropriate metadata.  I believe the fewer field tags required leads to a more maintainable codebase for developers.

**Describe the solution you'd like**
I believe most developers of Golang applications that use model objects for a RESTful API are very comfortable supplying the `json:` field tag.  The `omitempty` label becomes an interesting/powerful signal to determine if an attribute is `required` (or not) - without having to specifying additional `binding` and/or `validate` tags.

I would like for the `IsRequired` function in `field_parser.go` to treat the `omitempty` label of the `json` tag as `required: false` for the purposes of Swagger generation.  This behavior, in combination with the `--requiredByDefault` flag, can allow developers to have a reliable `required[]` attribute defined on their endpoints without needing to provide additional tags to manage the `required` property.

**Describe alternatives you've considered**
I have evaluated using `binding` and/or `validate` tag to provide appropriate metadata to control the `required` attribute of the Swagger specification - but didn't love the implications of that approach as it would require dozens upon dozens of changes to properly decorate the models. Additionally, the changes felt redundant as it was really mirroring the structure of the `omitempty` label.

**Additional context**
I have really tried to scratch my head and _have a think_ with respect to how/why treating `omitempty` as `required: false` could be dangerous/misleading - but have not come up with any compelling evidence.  `omitempty` - by definition - implies the attribute in question **could be empty** - and if a given attribute **could be empty** - it seems tautological then that the attribute in question **is not required**.  But I am certainly open to the insights and opinions of the community if I am being dense and overlooking any obvious pitfalls here!

Additionally, while certainly I want and will honor any discussions brought forth on this issue - I did fork this repo and create a preliminary commit to "see" this solution in action.  I believe it fits nicely and cleanly into the current codebase - and and confirm that this added functionality does in fact achieve my desired outcome of "simpler" `tag` management - while exposing appropriate `required` details within the Swagger:
- https://github.com/swaggo/swag/compare/master...andyatmiami:swag:feat/omitempty-as-optional?expand=1

sdghchj:
released

PR #2041, opened 2025-07-15
feat: json:omitempty marks field as optional

andyatmiami:
**Describe the PR**
- Enhances field parser to recognize 'omitempty' in `json:` tag as `required: false`, updating IsRequired logic (following existing pattern present in parsing logic) and adding corresponding test case. 
- Update README to document the new behavior w.r.t `json:omitempty`

**Relation issue**
Related: #2040  

**Additional context**
While I believe this behavior makes the most sense to use in conjunction with the `--requiredByDefault` flag - there is no hard requirement or dependency to force both these options to co-exist.  Implementation gives preference to existing supported `tag` values ( `binding` + `validate`) so anything presently relying on them for `required` calculation is unaffected by this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -541,6 +541,7 @@ type Foo struct {
 Field Name | Type | Description
 ---|:---:|---
 <a name="validate"></a>validate | `string` | 	Determines the validation for the parameter. Possible values are: `required,optional`.
+<a name="json"></a>json | `string` | JSON tag options. The `omitempty` option will mark the field as not required.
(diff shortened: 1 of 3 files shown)


sdghchj (approved):

ant1g:
Any chance this can get prioritized somehow? I can propose the same PR against the v2 branch? Thanks!

sdghchj:
@ant1g PR is welcome.

PR #2083, opened 2025-10-12
fix: generic types name generation

HermanPlay:
**Describe the PR**
During `swag init` when using `useStructName` flag. Fixed the issue where types using in generic were generated with full path instead of only type name.

**Relation issue**
Expands changes introduced in https://github.com/swaggo/swag/pull/2043

**Additional context**
Assume following code example:
```
import (
	"github.com/mini-maxit/backend/internal/api/http/httputils"
	"github.com/mini-maxit/backend/package/domain/schemas"
)

// @Success      200 {object} httputils.APIResponse[schemas.WorkerStatus]
func foo() {}
```

When generating swagger with `swag init --st --dir . --ot yaml --st`
The resulting type definition in swagger looked something like:
```
APIResponse-github_com_mini-maxit_backend_package_domain_schemas_WorkerStatus:
  properties:
    data:
      $ref: '#/definitions/WorkerStatus'
    ok:
      type: boolean
  type: object
```
We see that `APIResponse` is generated correctly with only struct name, when type used in generic is generated with full path.

After change the definion looks correct, with underling generic type shortened:
```
APIResponse-WorkerStatus:
  properties:
    data:
      $ref: '#/definitions/WorkerStatus'
    ok:
      type: boolean
  type: object
```

I know the tests are broken now. However, I cannot think of any better solution right, and help is welcomed and will be highly appreciated

### DIFF
--- a/generics.go
+++ b/generics.go
@@ -30,10 +30,6 @@ func (t *genericTypeSpec) TypeName() string {
 	return t.Name
 }
 
-func normalizeGenericTypeName(name string) string {
-	return strings.Replace(name, ".", "_", -1)
-}
-
 func (pkgDefs *PackagesDefinitions) getTypeFromGenericParam(genericParam string, file *ast.File) (typeSpecDef *TypeSpecDef) {
 	if strings.HasPrefix(genericParam, "[]") {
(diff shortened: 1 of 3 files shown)


sdghchj:
Some tests failed.

PR #2101, opened 2025-11-20
perf: add --parseGoPackages flag to speed up parse source

okhowang:
**Describe the PR**
using go modern packages API which can make use of cache and multicore

**Relation issue**
#2097

**Additional context**
there is some different with old behaviour.
1. packages API will not parse go files filtered by build tag.
2. packages API will failed when source file couldn't be go build.
3. packages API will failed if all search dir are not in same go module.

on my project `time swag init --parseGoPackages`
```
real    0m5.026s
user    0m48.465s
sys     0m6.592s
```
compare with `time swag init`
```
real    0m26.623s
user    1m24.083s
sys     0m32.631s
```

we can make use of packages API to simplify resolving constant variable further


sdghchj (approved):

sdghchj:
sorry, I see, it OK

Issue #2114, opened 2025-12-11
Generation panic on custom string


nesty156:
**Describe the bug**
When running the swag v2 with this flags: `swag init -v3.1 --parseDependency --parseInternal --parseDepth 1 -g main.go` I get panic because of custom string

```go
type ThreatType string

const (
	MALWARE               ThreatType = "MALWARE"
	SPYWARE               ThreatType = "SPYWARE"
)

type Threat struct {
	ThreatType ThreatType `json:"threatType" form:"threatType"`
}
```

**To Reproduce**
1. Install swag v2
2. Run the docs generation: `swag init -v3.1 --parseDependency --parseInternal --parseDepth 1 -g main.go`

**Expected behavior**
No panic :)

**Screenshots**

**Your swag version**
e.g. 2.0.0

**Your go version**
e.g. 1.25.0

**Desktop (please complete the following information):**
 - OS: macOS 15.7.2

**Additional context**
This issue https://github.com/swaggo/swag/issues/2078 has the same problem

Issue #2078, opened 2025-09-26
Generation panic on custom type


mr1cloud:
**Describe the bug**
When running the Swagger documentation generation (swag init), the process fails with a panic: runtime error: invalid memory address or nil pointer dereference. The error occurs during schema generation for the model.ClubException struct, specifically when parsing the field:
```golang
type Date struct{ time.Time } // Custom type from base time.Time

type MyStruct {
    Dates []types.Date `json:"dates" db:"dates" validate:"dive,date" swaggertype:"array,string" format:"date"`
}
```

**To Reproduce**
1. Install swag v2
2. Run the docs generation task: go tool swag init -g internal/controller/rest/server.go -o docs/swagger --v3.1 (in my case)

**Expected behavior**
```yaml
model.MyModel:
  properties:
    dates:
      items:
        type: string
        format: date
      example:
        - "2025-01-01"
      type: array
      uniqueItems: false
```

**Screenshots**
N/A

**Your swag version**
2.0.0

**Your go version**
1.24.0

**Desktop (please complete the following information):**
 - OS: Windows 10

**Additional context**
Logs:
<details>

```log
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x0 pc=0xf49dd1]

goroutine 1 [running]:
github.com/swaggo/swag/v2.(*Parser).GetSchemaTypePathV3(0xc00029a000, 0xc0003850a0, 0x2)
        C:/Users/maxim/go/pkg/mod/github.com/swaggo/swag/v2@v2.0.0-rc4/parserv3.go:1065 +0x111
github.com/swaggo/swag/v2.(*tagBaseFieldParserV3).ComplementSchema(0xc0001be210, 0xc000350f00?)
        C:/Users/maxim/go/pkg/mod/github.com/swaggo/swag/v2@v2.0.0-rc4/field_parserv3.go:127 +0xae
github.com/swaggo/swag/v2.(*Parser).parseStructFieldV3(0xc00029a000, 0xc000350f00, 0xc0003fe540)
        C:/Users/maxim/go/pkg/mod/github.com/swaggo/swag/v2@v2.0.0-rc4/parserv3.go:989 +0x3b8
github.com/swaggo/swag/v2.(*Parser).parseStructV3(0xc00029a000, 0xc000350f00, 0xc0003f8750)
[...]

PR #2115, opened 2025-12-11
fix: prevent panic when parsing custom string types in OpenAPI v3

nesty156:
**Describe the PR**
Prevent panic when parsing custom string types with form tags in OpenAPI v3

**Relation issue**
Fixing https://github.com/swaggo/swag/issues/2114 and probably https://github.com/swaggo/swag/issues/2078

### DIFF
--- a/parserv3.go
+++ b/parserv3.go
@@ -1048,7 +1048,7 @@ func (p *Parser) parseStructFieldV3(file *ast.File, field *ast.Field) (map[strin
 		tagRequired = append(tagRequired, fieldName)
 	}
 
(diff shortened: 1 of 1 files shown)


shindonghwi (commented):
The fix looks good! Adding nil checks for `schema` and `schema.Spec` before accessing `Extensions` prevents the panic.

One minor suggestion: It might be worth adding a unit test to cover this edge case (custom string types with form tags), to prevent regression in the future.

btw, just a small suggestion about the test - no big deal if you skip it

PR #2117, opened 2025-12-16
fix: swag fmt not working with relative path like ../

### LINKED ISSUE #2112: Swag Fmt ignores --dir option
**Describe the bug**
`swag fmt --dir ../` does not format files in `../**`.
It works when doing `cd ../ && swag fmt`.

**To Reproduce**
In a go project, go into any directory and try the command. (e.g., in cmd/, pkg/, ...).

**Expected behavior**
All files recursively down from the specified directory are formatted.

**Your swag version**
1.16.6

**Your go version**
go1.25.3 linux/amd64

 - OS: fedora linux
 - Version: 43

**Usecase**
We have the entry point to our API in cmd/main.go. In that file we have `//go:generate` commands to generate the swagger doc, as well as to format the files. We have noticed that the files in the repo are not formatted.

shindonghwi:
## Summary
- Fix `swag fmt --dir ../` not formatting files when using relative parent path
- The `excludeDir` function incorrectly treated `..` as a hidden folder because it checked if path starts with `.` and has length > 1

## Changes
- Modified hidden folder detection to exclude `..` from being treated as hidden
- Added test case for relative path formatting

Fixes #2112

### DIFF
--- a/format/format.go
+++ b/format/format.go
@@ -87,9 +87,10 @@ func (f *Format) Build(config *Config) error {
 }
 
 func (f *Format) excludeDir(path string) bool {
-	return f.exclude[path] ||
-		filepath.Base(path)[0] == '.' &&
-			len(filepath.Base(path)) > 1 // exclude hidden folders
+	base := filepath.Base(path)
+	// exclude hidden folders (starting with .) but not "." or ".."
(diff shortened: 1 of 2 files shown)


CodingTil:
Push

PR #2131, opened 2026-01-10
fix: avoid oneOf when @Accept adds default schema

taba2424:
**Describe the PR**

- Fix OpenAPI v3 requestBody schema generation so the default schema created by `@Accept` does not get merged via `oneOf` with the actual body schema.
- Replace the placeholder `type: object` with the real body schema when `@Param ... body` is parsed after `@Accept`.
- Add a regression test covering `@Accept json` + body parameter.

**Relation issue**

N/A (no related issue)

**Additional context**

Example setup:

```go
// @Accept json
// @Param request body types.CreateRequest true "作成リクエスト"
```

Command (example):

Generated specs using swag init with OpenAPI v3.1 output and YAML format.
Before, the generated requestBody schema included:

```yaml
oneOf:
  - type: object
  - $ref: '#/components/schemas/CreateRequest'
```

This PR ensures the placeholder type: object from @Accept is replaced by the actual body schema, resulting in only the $ref entry.

### DIFF
--- a/operationv3.go
+++ b/operationv3.go
@@ -6,6 +6,7 @@ import (
 	"go/ast"
 	"log"
 	"net/http"
+	"reflect"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


kundkingan (approved):

ubogdan:
@taba2424 Thanks for your contribution.

kundkingan:
When can we expect a new rc release @ubogdan

PR #2148, opened 2026-03-11
feat: support override attributes and type placeholders in .swaggo

### LINKED ISSUE #1852: Support swagger type format for global override
I am trying to declare a global override in .swaggo to replace my custom date struct with the string swagger type using the date-time format 

The solution I would be along the lines of an additional formatting parameter after the type in the 'replace' function e.g. 

```replace src/types.NullTime string date-time```

One workaround I've explored is to replace this type with another custom type with the correct swagger format:
```
type NullTimeSwagger struct {
	String *string `swaggertype:"string" format:"date-time"`
}
```
However, the generated doc places the date-time string inside a struct, which is undesirable.

### LINKED ISSUE #2147: Support attributes (nullable, format, optional) and type placeholders in .swaggo overrides
## Problem

The `.swaggo` override file lets users globally replace types (e.g., `replace sql.NullString string`), but falls short when the replacement needs additional schema metadata — specifically **nullable**, **optional**, and **format** properties.

### Use case: three-state PATCH fields

In REST APIs, PATCH endpoints often need to distinguish three states per field:

| State | JSON | Meaning |
|-------|------|---------|
| **Absent** | field not in payload | Don't change |
| **Null** | `"field": null` | Clear the value |
| **Value** | `"field": 42` | Set the value |

A common Go solution is a generic `Optional[T]` with custom `UnmarshalJSON`. For correct OpenAPI output, each field using this type should be **optional** (not in `required`) and **nullable** (`x-nullable: true`).

Today, users must annotate **every field** with struct tags:

```go
type PatchUserRequest struct {
    Name Optional[string] `json:"name" validate:"optional" extensions:"x-nullable"`
    Bio  Optional[string] `json:"bio" validate:"optional" extensions:"x-nullable"`
}
```

This defeats the purpose of centralized overrides — the semantics are properties of the *type*, not individual fields.

### No generic catch-all

Each `Optional[T]` instantiation must be listed individually:

```
replace common.Optional[string] string
replace common.Optional[int] integer
replace common.Optional[models.Address] models.Address
// ... one line per type
```

### Related

This also addresses #1852 — supporting `format` in global overrides (e.g., `date-time` for time types).

### 1. Override attributes (`key:value` syntax)

Extend `replace` to accept `key:value` attributes after the replacement type:

```
replace common.Optional[string] string optional:true nullable:true
replace common.Optional[time.Time] string optional:true nullable:true format:date-time
```

Supported attributes:
- `optional:true` — removes the field from `required`
- `nullable:true` — adds `x-nullable: true`
- `format:<value>` — sets `format` (e.g., `date-time`, `uuid`)

### 2. Type parameter placeholder (`$T`)

Allow `$T` in source and replacement types for generic catch-all rules:

```
replace common.Optional[$T] $T optional:true nullable:true
```

When swag encounters `common.Optional[models.Address]`, it matches the pattern, captures `models.Address`, substitutes `$T`, resolves normally, and applies the attributes.

Exact matches take priority over placeholder patterns.

### Complete `.swaggo` example

```
replace common.Optional[string] string optional:true nullable:true
replace common.Optional[float64] number optional:true nullable:true
replace common.Optional[time.Time] string optional:true nullable:true format:date-time
replace common.Optional[$T] $T optional:true nullable:true
```

## Backward Compatibility

- **`.swaggo` format**: Fully backward compatible — existing lines work unchanged.
- **`SetOverrides` API**: Changes from `map[string]string` to `map[string]Override`. This is a breaking change for programmatic callers of `SetOverrides`, but this function is primarily used internally by the `gen` package.

## Open question: attribute naming convention

The current implementation uses clean, short attribute names:

```
replace common.Optional[string] string optional:true nullable:true format:date-time
```

An alternative would be to mirror the existing struct tag names more closely:

```
replace common.Optional[string] string validate:optional extensions:x-nullable format:date-time
```

The struct-tag-aligned syntax would be more familiar to existing swag users (`extensions:"x-nullable"` and `validate:"optional"` are already well-known), but the shorter names are arguably cleaner for a config file context. Happy to adjust either way — would appreciate maintainer input on the preferred convention.

## Implementation

I have a working implementation with tests ready as a PR: override struct, `key:value` parsing, `$T` matching, and attribute application. All existing tests pass.

isasmendiagus:
## Summary

- Add `key:value` attribute syntax to `.swaggo` `replace` directives (nullable, optional, format)
- Add `$T` type parameter placeholder for generic catch-all overrides
- Introduce `Override` struct to replace raw string type mappings

Closes #2147
Closes #1852

## Changes

- **`parser.go`** — `Override` struct, `matchOverride()`, `getOverrideForType()`, `applyOverrideAttrs()`, updated `getTypeSchema()` and `parseStructField()`
- **`gen/gen.go`** — `parseOverrides()` with `key:value` attribute parsing
- **`parser_test.go`** — new tests for nullable, format, placeholder, and `matchOverride`
- **`gen/gen_test.go`** — new test cases for attrs, format, malformed input, and placeholder
- **`generics_test.go`** — updated to use `Override` struct

## Usage

```
// Primitives with attributes
replace common.Optional[string] string optional:true nullable:true
replace common.Optional[time.Time] string optional:true nullable:true format:date-time

// Generic catch-all
replace common.Optional[$T] $T optional:true nullable:true
```

### DIFF
--- a/gen/gen.go
+++ b/gen/gen.go
@@ -183,7 +183,7 @@ func (g *Gen) Build(config *Config) error {
 		config.RightTemplateDelim = "}}"
 	}
 
-	var overrides map[string]string
+	var overrides map[string]swag.Override
 
 	if config.OverridesFile != "" {
 		overridesFile, err := open(config.OverridesFile)
@@ -379,8 +379,8 @@ func (g *Gen) formatSource(src []byte) []byte {
 }
 
 // Read and parse the overrides file.
-func parseOverrides(r io.Reader) (map[string]string, error) {
-	overrides := make(map[string]string)
+func parseOverrides(r io.Reader) (map[string]swag.Override, error) {
+	overrides := make(map[string]swag.Override)
 	scanner := bufio.NewScanner(r)
 
 	for scanner.Scan() {
@@ -393,24 +393,31 @@ func parseOverrides(r io.Reader) (map[string]string, error) {
 
 		parts := strings.Fields(line)
 
-		switch len(parts) {
-		case 0:
-			// only whitespace
+		if len(parts) == 0 {
 			continue
-		case 2:
-			// either a skip or malformed
-			if parts[0] != "skip" {
+		}
+
+		switch parts[0] {
+		case "skip":
+			if len(parts) != 2 {
+				return nil, fmt.Errorf("could not parse override: '%s'", line)
+			}
+			overrides[parts[1]] = swag.Override{}
+		case "replace":
+			if len(parts) < 3 {
 				return nil, fmt.Errorf("could not parse override: '%s'", line)
 			}
 
-			overrides[parts[1]] = ""
-		case 3:
-			// either a replace or malformed
-			if parts[0] != "replace" {
-				return nil, fmt.Errorf("could not parse override: '%s'", line)
+			attrs := make(map[string]string)
+			for _, attr := range parts[3:] {
+				kv := strings.SplitN(attr, ":", 2)
+				if len(kv) != 2 {
+					return nil, fmt.Errorf("malformed attribute '%s' in override: '%s'", attr, line)
+				}
+				attrs[kv[0]] = kv[1]
 			}
 
-			overrides[parts[1]] = parts[2]
+			overrides[parts[1]] = swag.Override{Type: parts[2], Attrs: attrs}
 		default:
 			return nil, fmt.Errorf("could not parse override: '%s'", line)
 		}
--- a/gen/gen_test.go
+++ b/gen/gen_test.go
@@ -684,43 +684,43 @@ func TestGen_parseOverrides(t *testing.T) {
 	testCases := []struct {
 		Name          string
 		Data          string
-		Expected      map[string]string
+		Expected      map[string]swag.Override
 		ExpectedError error
 	}{
 		{
 			Name: "replace",
 			Data: `replace github.com/foo/bar baz`,
-			Expected: map[string]string{
-				"github.com/foo/bar": "baz",
+			Expected: map[string]swag.Override{
+				"github.com/foo/bar": {Type: "baz", Attrs: map[string]string{}},
 			},
 		},
 		{
 			Name: "skip",
 			Data: `skip github.com/foo/bar`,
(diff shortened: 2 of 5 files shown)


maxclaus:
@sdghchj could you please check this one out when you have some time? It would be a great addition to the tool, being able to handle generic wrapper types globally is something a lot of codebases need.

I tested locally on my project and confirmed it works great. Here's a concrete example of what it unlocks.

Context: we use a tri-state optional.Value[T] wrapper for PATCH endpoints (absent / null / present), since Go's JSON marshaller does not distinguish null from absent (undefined) fields.

Before, we had to maintain a parallel `*Doc` struct purely for swagger:

```go
// PatchRequestPayloadDoc is a doc-only type used to generate correct swagger output.
// It mirrors PatchRequestPayload but replaces optional.Value[T] with *T.
type PatchRequestPayloadDoc struct {
    Date             *int64            `json:"date"`
    Status           *domain.Status    `json:"status"`
} // @name PatchRequestPayload

type PatchRequestPayload struct {
    Date             optional.Value[int64]            `json:"date,omitempty"`
    Status           optional.Value[domain.Status]    `json:"status,omitempty"`
}

...
// @Param body body PatchRequestPayloadDoc true "Patch payload"
```

After, with a single line in `.swaggo`:
```
replace optional.Value[$T] $T nullable:true
```
The real struct is used directly in the annotation, no duplicate needed:
```go
type PatchRequestPayload struct {
    Date             optional.Value[int64]            `json:"date,omitempty"`
    Status           optional.Value[domain.Status]    `json:"status,omitempty"`
}

...
// @Param body body PatchRequestPayload true "Patch payload"
```

Generated output renders each field as its inner type with `x-nullable: true`, exactly correct. Zero `optional.Value-*` intermediate definitions in the spec.

teodor-pripoae:
Any update ? We are also looking into using optionals with swag.

alexjplant:
^ ditto. This would be a real boon for `PATCH` endpoints.

isasmendiagus:
Hi @sdghchj, could you please take a look ?

Issue #2150, opened 2026-03-13
Add Discriminator annotation support for OpenAPI v3


lukasab:
**Is your feature request related to a problem? Please describe.**

When using oneOf polymorphism in OpenAPI v3 (multiple `@Sucess` annotations with the same status code), there is no way to declare a [discriminator object](https://spec.openapis.org/oas/v3.0.3#discriminator-object). Clients that need to deserialize polymorphic responses have no machine-readable hint about which property identifies the concrete type.

**Describe the solution you'd like**

A new `@Discriminator` annotation for route handlers that maps onto the OpenAPI 3.0 discriminator object (`propertyName` + optional `mapping`):

```go
// With mapping:
// @Discriminator pet_type cat=#/components/schemas/web.Cat,dog=#/components/schemas/web.Dog
// @Success 200 {object} web.Cat
// @Success 200 {object} web.Dog
// @Router /pets/{id} [get]

// Property name only:
// @Discriminator pet_type
// @Success 200 {object} web.Cat
// @Success 200 {object} web.Dog
// @Router /pets/{id}/no-mapping [get]
```

Expected generated output:
```yaml
responses:
  '200':
    content:
      application/json:
        schema:
          oneOf:
            - $ref: '#/components/schemas/web.Cat'
            - $ref: '#/components/schemas/web.Dog'
          discriminator:
            propertyName: pet_type
            mapping:
              cat: '#/components/schemas/web.Cat'
              dog: '#/components/schemas/web.Dog'
```

**Describe alternatives you've considered**
- Manually editing the generated spec after the fact — fragile and breaks on regeneration.
- Using a struct-level oneOf tag with a discriminator — not possible at the route level where the polymorphism is declared.

NiJeTi:
@lukasab First of all, thanks for the amazing job! Could you please either navigate how to implement or event contribute yourself a way to declare a nested discriminator (e.g. the body of my response has a nested array of objects which can be represented in `oneOf` manner and be discriminated by another property).

lukasab:
> [@lukasab](https://github.com/lukasab) First of all, thanks for the amazing job! Could you please either navigate how to implement or event contribute yourself a way to declare a nested discriminator (e.g. the body of my response has a nested array of objects which can be represented in `oneOf` manner and be discriminated by another property).

Thanks! Glad it's useful.

[...]

PR #1870, opened 2024-08-13
Feat openapiv3 oneof

KKKIIO:
**Describe the PR**
Adds support for the oneOf feature in OpenAPI v3, allowing you to describe a field/response that can have multiple types.

**Relation issue**
https://github.com/swaggo/swag/issues/1363
https://github.com/swaggo/swag/issues/1617

**Additional context**
Unfortunately, the PR https://github.com/swaggo/swag/pull/1671 was reverted due to a compile error caused by the use of the `maps` package. 
The `maps` package was introduced in Go 1.21, but the project is currently using Go 1.18.

Issue #829, opened 2020-11-10
Composition and Inheritance (Polymorphism) support


denisvmedia:
Swagger 2.0 [supports](https://swagger.io/specification/v2/) polymorphism via `discriminator`:

```yaml
    definitions:
      Pet:
        type: object
        discriminator: petType
        properties:
          name:
            type: string
          petType:
            type: string
        required:
        - name
        - petType
      Cat:
        description: A representation of a cat
        allOf:
        - $ref: '#/definitions/Pet'
        - type: object
          properties:
            huntingSkill:
              type: string
              description: The measured skill for hunting
              default: lazy
              enum:
              - clueless
              - lazy
              - adventurous
              - aggressive
          required:
          - huntingSkill
      Dog:
        description: A representation of a dog
        allOf:
        - $ref: '#/definitions/Pet'
        - type: object
          properties:
            packSize:
              type: integer
              format: int32
              description: the size of the pack the dog is from
              default: 0
              minimum: 0
          required:
          - packSize
```

Does `swag` supports something similar? I currently defined it this way (of course, in production I have something more realistic, but as an example this works perfectly):

```
// @Param cat body model.PetWrap{data=jsonapi.Cat} true "Cat object"
// @Param dog body model.PetWrap{data=jsonapi.Dog} true "Dog object"
```

But this - although displays in Swagger UI - leads to an invalid Swagger document (multiple body params). So, is it somehow achievable?

ubogdan:
Yes, it can be done if you implement it and make the community happy with a PR.

PR #2151, opened 2026-03-13
Add support for discriminator in OpenAPI v3

lukasab:
**Describe the PR**
Adds a new `@Discriminator` route-handler annotation for OpenAPI v3. When a response (or request body) uses `oneOf` polymorphism (multiple `@Success` annotations with the same status code), a discriminator object can now be declared to tell clients which concrete type to use at runtime.

**Relation issue**
#2150 #829 

**Additional context**
Builds on the existing `oneOf` support introduced in #1870 and extended in #2059. The discriminator should only be applied to schemas that already contain a `oneOf` composition, as per the OpenAPI 3.0 spec.

**Implementation notes**
- Follows the same deferred-processing pattern as `@Produce`
- `discriminatorInfo *parsedDiscriminator` is stored on `OperationV3` to bridge the two phases. Open to feedback on whether there's a better place for this — it felt slightly out of place on the struct but couldn't find a cleaner alternative given the parse ordering constraints.
- Declaring `@Discriminator` twice on the same operation returns an error rather than silently overwriting.
- Applied to both responses and request body content.

### DIFF
--- a/operationv3.go
+++ b/operationv3.go
@@ -13,6 +13,11 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+type parsedDiscriminator struct {
+	propertyName string
+	mapping      map[string]string // nil when not specified
+}
+
 // OperationV3 describes a single API operation on a path.
 // For more information: https://github.com/swaggo/swag#api-operation
 type OperationV3 struct {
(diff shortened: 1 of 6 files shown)


ubogdan (approved):
LGTM

ubogdan:
@lukasab Thanks for your contribution.

kundkingan:
@ubogdan New release coming out soon?

PR #2164, opened 2026-05-04
feat: add --requiredByDefaultMode=pointer to mark non-pointer fields as required

### LINKED ISSUE #901: Is it possible to mark every non pointer field as required?
Hey, I would like swaggo to generate my swagger.json file so that every field which is not a pointer is marked as required. I know that you can add the `validate:"required` tag, but I would prefer to not have to do this explicitly everywhere.

Is it possible to do this?

chiptus:
Adds a new --requiredByDefaultMode flag to control which fields are marked as required by default. Supported values are "all" (all fields required) and "pointer" (only non-pointer fields required), reflecting Go semantics where non-pointer fields always have a value and cannot be nil.

Explicit tags (binding/validate:"optional", json:",omitempty") still take precedence. The existing --requiredByDefault flag is unchanged for full backward compatibility.

fix #901

### DIFF
--- a/cmd/swag/main.go
+++ b/cmd/swag/main.go
@@ -15,37 +15,37 @@ import (
 )
 
 const (
-	searchDirFlag            = "dir"
-	excludeFlag              = "exclude"
-	generalInfoFlag          = "generalInfo"
-	pipeFlag                 = "pipe"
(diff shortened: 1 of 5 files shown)


nxvinh22:
Someone please review

PR #2174, opened 2026-06-04
Respect parseDepth with go list dependencies

### LINKED ISSUE #1269: No longer respects --parseDepth when processing dependency tree
**Describe the bug**
Versions 1.8.3 and higher default to using the "go list" package listing and the new parser.getAllGoFileInfoFromDepsByList() function. This newer code does not reference the setting for --parseDepth. While the behavior can be overridden for now by setting --parseGoList=false, causing swag to run the old parser.getAllGoFileInfoFromDeps() function (which uses a depth tree and _does_ respect --parseDepth), it's still misleading and regressive behavior.

**Expected behavior**
If I set a non-default value for --parseDepth, I expect swag to honor it.

**Your swag version**
1.8.3

**Your go version**
e.g. 1.17.6

 - OS: iOS
 - Browser: chrome
 - Version: 103.0.5060.114

puneetdixit200:
**Describe the PR**
Fixes the `ParseUsingGoList(true)` dependency path so `parseDepth` limits how far dependency packages are parsed. The go-list command still gathers dependency metadata, then filters packages by their distance from the root package before parsing source files.

**Relation issue**
Fixes https://github.com/swaggo/swag/issues/1269

**Additional context**
Added a regression fixture where the app imports a direct dependency that imports a transitive dependency. The tests verify depth 1 excludes the transitive package and depth 2 includes it.

Tests:
- `GOMODCACHE=/tmp/swag-1269-gomodcache GOCACHE=/tmp/swag-1269-gocache go test . -run 'Test(FilterPackagesByParseDepth|ParseGoListRespectsParseDepth|ParseGoList)' -count=1`\n- `GOMODCACHE=/tmp/swag-1269-gomodcache GOCACHE=/tmp/swag-1269-gocache go test ./... -count=1`\n- `git diff --check`


sharunkumar:
I tried this implementation with a very large internal repo, and the performance wins that we got earlier with `parseGoList=true` were negated by this fix. I have an alternate implementation for fixing `parseDepth` which I can put up shortly

PR #2175, opened 2026-06-11
fix: respect parseDepth for operations in the go list dependency walk

### LINKED ISSUE #1269: No longer respects --parseDepth when processing dependency tree

sharunkumar:
**Describe the PR**
When `parseGoList=true` was specified, `parseDepth` was no longer being respected. This PR fixes that in a backwards compatible way

benchmarked for speed & correctness (comparing swagger.json before/after) on a large internal repository

**Related issue**
Fixes https://github.com/swaggo/swag/issues/1269

**Additional Context**

https://github.com/swaggo/swag/pull/2174#issuecomment-4685438255

### DIFF
--- a/cmd/swag/main.go
+++ b/cmd/swag/main.go
@@ -267,6 +267,7 @@ func initAction(ctx *cli.Context) error {
 		RequiredByDefault:   ctx.Bool(requiredByDefaultFlag),
 		CodeExampleFilesDir: ctx.String(codeExampleFilesFlag),
 		ParseDepth:          ctx.Int(parseDepthFlag),
+		ParseDepthSet:       ctx.IsSet(parseDepthFlag),
 		InstanceName:        ctx.String(instanceNameFlag),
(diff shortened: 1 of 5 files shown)


sharunkumar:
cc @sdghchj

PR #2189, opened 2026-07-03
fix(oas3): remove Swagger 2 schemes field from runtime docs template

plusiv:
**Describe the PR**
This PR fixes OpenAPI 3 runtime docs generation.

The static OpenAPI 3 JSON can be valid, but the docs generated through `swag.ReadDoc(...)` can include:

```
"schemes": null
```

`schemes` is a Swagger 2 field and is not valid in OpenAPI 3. OpenAPI 3 should use servers instead.

This change removes schemes from the OpenAPI 3 runtime template. Swagger 2 behavior is not changed.

Also added tests to detect future regressions.

**Relation issue**
N/A

**Additional context**
Any strict OpenAPI 3 validator can fail because this field is not part of the OpenAPI 3 spec. Figured out about it because of the Mintlify OpenAPI 3 validator


plusiv:
@ubogdan @easonlin404 wdyt?

PR #2200, opened 2026-07-19
fix(v3): build swaggertype array overrides with items, not additional…

kklem0:
Fixes the v3 half of the `swaggertype:"array,..."` override (see umbrella issue).

`BuildCustomSchemaV3`'s `ARRAY` case stored the element schema in `additionalProperties` (a map keyword) and left `Items` nil — the line was marked `// TODO: check if this is correct`. Any field using an array swaggertype override then panicked in `complementSchema` dereferencing `schema.Items.Schema.Spec`:

```go
AuthorIDs []pgtype.Int4 `json:"author_ids" swaggertype:"array,integer"`
```

```
panic: runtime error: invalid memory address or nil pointer dereference
  github.com/swaggo/swag/v2.(*tagBaseFieldParserV3).complementSchema
```

This mirrors how `parseTypeExprV3` builds schemas for ordinary Go slices (`Items = spec.NewBoolOrSchema(false, itemSchema)`).

After the fix the override renders as expected:

```json
{"type": "array", "items": {"type": "integer"}}
```

Adds `TestBuildCustomSchemaV3` covering array/object/primitive/error cases. Validated on a production codebase (126 operations): generation completes and array overrides render correctly.


YoungMind1:
This fixes the panic.

schemav3_test.go
YoungMind1, line 1: I think it would be more desirable that you would mirror `shcema_test.go` tests too.

PR #2201, opened 2026-07-19
fix(v3): only skip fields with an explicit json:"-", not any missing …

kklem0:
Fixes the silent-empty-schema bug for structs tagged only for other encoders (see umbrella issue).

`ShouldSkip` treated an empty `JsonName()` as "skip this field", which conflates two different situations: an explicit `json:"-"` and a field that has a struct tag without any `json:` key (xml/yaml/validate/...). encoding/json still serializes the latter under the Go field name, and `FieldName()` already resolves such fields via the property naming strategy — but they never got that far, so structs like:

```go
type BookTei struct {
    XMLName   xml.Name  `xml:"http://www.tei-c.org/ns/1.0 TEI"`
    TeiHeader TeiHeader `xml:"teiHeader"`
    Text      TeiText   `xml:"text"`
}
```

were emitted as `{"type": "object"}` with no properties and no warning, and every type referenced only through them vanished from the spec (15 schemas in our codebase). swag v1 documents the same struct as `{teiHeader, text, xmlname}`; this PR restores that behavior by only skipping an explicit `json:"-"`.

Adds `TestShouldSkipV3`. Validated on a production codebase: schema count returns to parity with the v1 generator (130/130).


YoungMind1:
Duplicate of #2141

sdghchj (approved):

Issue #2197, opened 2026-07-14
v2: every @Param body request body wrapped in a spurious 2-branch oneOf with a $ref sibling


ahridin:
**swag version:** v2.0.0-rc5

For a simple, single-type request body:

```go
// @Param body body MyRequest true "request payload"
```

the generated `requestBody.content["application/json"].schema` is emitted as:

```json
{ "oneOf": [ { "type": "object" }, { "$ref": "#/components/schemas/MyRequest", "description": "request payload" } ] }
```

instead of the expected plain:

```json
{ "$ref": "#/components/schemas/MyRequest" }
```

This causes two separate problems downstream:

1. The `$ref` sibling (`description` next to `$ref`) is invalid per OAS 3.0 (`$ref` siblings are ignored/disallowed) and rejected by strict validators like kin-openapi even under 3.1.
2. Because it's a genuine 2-branch `oneOf` rather than a single reference, `openapi-generator` can't dedupe it back to the named schema — it synthesizes a throwaway per-operation wrapper model (e.g. `my_operation_request.py`) instead of reusing `MyRequest`, for every single endpoint with a body parameter.

**Expected:** a single-type `@Param body` should emit a plain `$ref` to the schema, not a 2-branch `oneOf`.

klajok:
This is a regression. It used to work fine in v2.0.0-rc4.

PR #2210, opened 2026-08-05
fix: avoid oneOf when body is an array

klajok:
~**Describe the PR**
Fix regression introduced in `v2.0.0-rc5` that body of the array type generates `oneOf` structure in schema.~

~**Relation issue**
https://github.com/swaggo/swag/issues/2197~

~**Additional context**
Commit 13ced53b02e7754c5d11fcd6146fb472a91d7f48 introduced multi-body-param `oneOf` support by rewriting [`fillRequestBody`](operationv3.go:480). The merge logic used this condition to decide whether to create a `oneOf`:~
```go
} else if mediaType.Spec.Schema.Ref != nil || mediaType.Spec.Schema.Spec.OneOf == nil {
    // wrap in oneOf
```
~This fired for **every** call where a schema was already set — including the very first `@Param body` call when `@Accept json` had pre-populated the content map with a default `{type: "object"}` schema. The `isDefaultAcceptSchema` helper added by a later fix (#2131) only partially addressed this; it still failed for array schemas (`[]MyType`) because arrays are not "default accept schemas".~

~Added a `requestBodyFilled bool` field to [`OperationV3`](operationv3.go:30) that is set to `true` on the **first** `fillRequestBody` call. This let us remove the now-unnecessary `isDefaultAcceptSchema` function and its `reflect` import entirely.~


klajok:
Tests are too simple. No proof that the fix is indeed correct. Closed (for now)

Issue #1949, opened 2024-12-18
[v2]: Support `examples` keywords


KevinThompsonYCRX:
**Is your feature request related to a problem? Please describe.**
The `example` keyword has been deprecated in OpenAPI 3.1 in favor of `examples`: 

**Describe the solution you'd like**
When the example tag is parsed, place it in a list under `examples`. Defining a valid separator between examples would be good, but it will also need to be escapable.

**Additional context**
Blog post for 3.1 - you'll need to scroll down to the "Use examples not example" section
https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0

AlexOConnorHub:
Is anyone working on this? I started fiddling around with examples for parameter's in the URL. For the annotation I used `@Param.examples <paramName> <exampleSlug> <exampleField> <exampleFieldValue>`, where `exampleField` can be `description`, `summary`, or `value`.

I saw that currently, `@Param` supports an `Example(<example>)` at the end of the line, but due to the fact that examples is an array and that items in the array can contain multiple fields, I think it should likely be broken out from just `@Params` so that the comment lines do not get too long for those looking to generate multiple examples.

I was able to get this annotation
```
// @Summary Returns user record by ID
// @Description Useful for finding a user's username
// @Tags Users
// @Produce json
// @Success 200 {object} map[string]interface{}
// @Failure 404 {object} map[string]string
// @Param uuid path string true "User ID"
// @Param.examples uuid user1 summary User 1
// @Param.examples uuid user1 This is the UUID of "User 1"
// @Param.examples uuid user1 value 019785fe-4eb4-766e-9c45-bec7780972a2
// @Param.examples uuid user2 summary User 2
// @Param.examples uuid user2 description This is the UUID of "User 2"
// @Param.examples uuid user2 value 019785fe-4eb4-766e-9c45-c1f83e7c1f1f
// @Router /user/{uuid} [get]
```

To lead to this in the docs
```
paths:
  /user/{uuid}:
    get:
      description: Useful for finding a user's username
      parameters:
      - description: User ID
        examples:
          user1:
            description: This is the UUID of "User 1"
            summary: User 1
            value: 019785fe-4eb4-766e-9c45-bec7780972a2
          user2:
            description: This is the UUID of "User 2"
            summary: User 2
            value: 019785fe-4eb4-766e-9c45-c1f83e7c1f1f
        in: path
        name: uuid
        required: true
        schema:
          type: string
      responses:
[...]

PR #2213, opened 2026-08-20
[v2] Add support for "examples" keyword

aitorcas23:
**Describe the PR**
Use "examples" instead of "example" tag.

"example" is now deprecated on openapi v3. This pull request comes from #1949. As mentioned in this [openapi migration guide](https://github.com/swaggo/swag/issues/1949) "examples" is now preferred over "example" in openapi v3.

This pull request only targets swapping "example" with "examples" and wrapping the existing example in a list. This pull request doesn't target support for multiple examples. So this is just for suppressing the existing deprecated warning.

**Relation issue**
https://github.com/swaggo/swag/issues/1949


YoungMind1 (approved):
Works as expected.

